PR #1861, opened 2014-10-29
Add NMap tool for getting IP.

yumingfei:
In some condition, vm may have fixed IP, but tcpdump couldn't get it until some packaged sent.
So can we start another tool 'nmap' parallel when tcpdump is working?
Following codes are simple description of this tool, just for hints. :)  

Signed-off-by: Yu Mingfei <yumingfei@cn.fujitsu.com>

### DIFF
--- a/virttest/utils_env.py
+++ b/virttest/utils_env.py
@@ -6,6 +6,7 @@ import re
 import time
 
 import utils_misc
+import utils_net
 import virt_vm
 import aexpect
 import remote
@@ -145,6 +146,7 @@ class Env(UserDict.IterableUserDict):
         empty = {"version": version}
         self._filename = filename
         self._tcpdump = None
+        self._nmap = None
         self._params = None
         self.save_lock = threading.RLock()
         if filename:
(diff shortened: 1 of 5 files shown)


yumingfei:
Updated to do not raise when nmap is not installed.

virttest/utils_env.py
```diff
@@ -369,3 +371,31 @@ def stop_tcpdump(self):
...
+            if iface in ifaces:
+                ifaces.remove(iface)
+        self._nmap.start_nmap(ifaces, timeout=None)
+
+    def start_nmap(self, params):
+        self._params = params
+        self._start_nmap()
+
+    def stop_nmap(self):
+        if self._namp is not None:
```
spcui: small typo, should be self._nmap

yumingfei: Thanks, update it soon.

virttest/virt_vm.py
```diff
@@ -1003,9 +1003,29 @@ def wait_for_login(self, nic_index=0, timeout=LOGIN_WAIT_TIMEOUT,
                 when remote login (ssh, rss) failed.
         :return: A ShellSession object.
         """
+        def _run_nmap(nmap, stop=False):
+            if stop:
+                nmap.stop_nmap()
+                return
+            nmap.start_nmap(utils_net.Bridge().list_br(), timeout=None)
```
ypu: The utils_net.Bridge will only can list the bridge that can be showed by 'brctl show' which means ovs bridge can not be listed by this function. So this will make this part failed when we run tests in a host using ovs bridge.

Based on this seems we need add the ovs bridge list to here too. And can we make this part also can use parameters: "nmap_interfaces"? It will save our time to get the mac addr and ip by using nmap.

yumingfei: Of course. The parameter 'nmap_interfaces' has already been used in utils_env.

yumingfei: I don't know how to add ovs bridge to my list.
So I leave it to nmap_interfaces in base.cfg, what do you think?

ypu: Just check the utils_net.py, seems we are leak of the function to list the brs in the host. Maybe it is better to add one to it. There are two global variable in util_net: __ovs and __bridge. They can be used directly to handle the different type of device init part.

lmr:
Hello @yumingfei, if you could be so kind, please rebase it and re-open it against

Thanks!

PR #1883, opened 2014-11-13
virt: Add catch_monitor for ScreenDump and VmRegister threads.

FengYang:
We have many thread need send monitor command. So Add a
catch_monitor for ScreenDump and VmRegister threads to reduce
acquire exclusive lock fail.

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -2635,14 +2635,14 @@ class VM(virt_vm.BaseVM):
     def monitor(self):
         """
         Return the main monitor object, selected by the parameter main_monitor.
-        If main_monitor isn't defined, return the first monitor.
-        If no monitors exist, or if main_monitor refers to a nonexistent
-        monitor, return None.
+        If main_monitor isn't defined or it refers to a nonexistent monitor,
+        return the first monitor.
+        If no monitors exist, return None.
         """
         for m in self.monitors:
             if m.name == self.params.get("main_monitor"):
                 return m
-        if self.monitors and not self.params.get("main_monitor"):
+        if self.monitors:
             return self.monitors[0]
         return None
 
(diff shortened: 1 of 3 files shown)


xutian:
Agree with @ypu's comments but it works to me, so give a ACK.

Thanks,
@xutian

ypu:
@FengYang seems still have a monitor command will shows up in the monitor log: query-status. And it is caused by the vm.is_alive() in screendump thread. Can we also make it in the catch_monitor log?

FengYang:
Hi @ypu will look into it.

virttest/qemu_vm.py
```diff
@@ -2642,6 +2646,22 @@ def monitor(self):
...
+        Return the catch monitor object, selected by the parameter
+        catch_monitor.
+        If catch_monitor isn't defined, return the last monitor.
+        If no monitors exist, or if catch_monitor refers to a nonexistent
+        monitor, return None.
+        """
+        for m in self.monitors:
+            if m.name == self.params.get("catch_monitor"):
+                return m
+        if self.monitors and not self.params.get("catch_monitor"):
```
xutian: Whatever happens, we still need a monitor to check vm status,  so It's better to return last monitor.

thanks,

@xutian

FengYang: On 02/02/2015 11:07 AM, Xu Tian wrote:
>
> In virttest/qemu_vm.py
> <https://github.com/autotest/virt-test/pull/1883#discussion_r23905374>:
>
> > @@ -2642,6 +2646,22 @@ def monitor(self):
> >              return self.monitors[0]
> >          return None
> >  
> > +    @property
> > +    def catch_monitor(self):
> > +        """
> > +        Return the catch monitor object, selected by the parameter
> > +        catch_monitor.
> > +        If catch_monitor isn't defined, return the last monitor.
> > +        If no monitors exist, or if catch_monitor refers to a nonexistent
> > +        monitor, return None.
> > +        """
> > +        for m in self.monitors:
> > +            if m.name == self.params.get("catch_monitor"):
> > +                return m
> > +        if self.monitors and not self.params.get("catch_monitor"):
>
> Whatever happens, we still need a monitor to check vm status, so It's
> better to return last monitor.
>
+ if self.monitors and not self.params.get("catch_monitor"):
+ return self.monitors[-1]

Sorry,  a little lost.

It will return last monitors when catch_monitor not available.

ypu:
ID: 1163690

ypu:
LGTM

xutian:
LGTM

Acked-by:  @xutian

ypu:
Thanks @FengYang and @xutian will merge it.

PR #1892, opened 2014-11-20
virttest.qemu_vm: make reboot work for linux serial session

xutian:
when reboot a linux guest  with a serial_login session, test will failed as "guest refuse to reboot", 
 root cause is we judge guest begin to reboot via checking serial not responsible.  


spcui:
Acked-by: @spcui

FengYang:
Thanks @xutian @spcui @ypu

PR #1901, opened 2014-11-26
Network xml update

rbian:


yumingfei:
OK, I have checked that only test "virtual_network.iface_options" calls it.
So it should have no effect on other tests.
Applied, thanks @rbian.

PR #1902, opened 2014-11-28
Update XMLElementList to support text element list

rbian:


cevich:
Oh gosh, I remember this stuff being fairly complex, I'll have to re-learn it first :grinning: let's see...

virttest/element_tree.py
```diff
@@ -455,10 +455,11 @@ def Element(tag, attrib={}, **extra):
 # @defreturn Element
 
 
-def SubElement(parent, tag, attrib={}, **extra):
+def SubElement(parent, tag, attrib={}, text=None, **extra):
```
cevich: Hrmmm, yeah I wouldn't do it this way.  This module was a port of ``xml.ElemenTree`` into virttest to isolate it from changes across various python versions.  Even the latest python 2.7 API does not contain this additional parameter.  It seems fine to add it now, but by using it, future compatibility will be broken.  This becomes important if autotest's python minimum version is updated and the custom ElementTree module is dropped in favor of native (to reduce maintenance and duplication).  It also breaks hard-to-find cases where the caller expects exactly four parameters in order.

virttest/libvirt_xml/accessors.py
```diff
@@ -796,9 +796,13 @@ def __call__(self, value):
...
                                                 str(item)))
                     raise xcepts.LibvirtXMLAccessorError(msg)
+                text = None
+                if len(element_tuple) == 3:
+                    text = element_tuple[2]
                 xml_utils.ElementTree.SubElement(parent,
                                                  element_tuple[0],
-                                                 element_tuple[1])
+                                                 element_tuple[1],
+                                                 text)
```
cevich: I don't remember specifically what this section does, but you could probably still have the same affect w/o modifying the ElementTree port.  All it costs is an extra line to update the ``text`` attribute, no?

cevich:
Okay, I understand now, yes this is probably a good/needed change.  My only concern is with it's form.  

IIRC, I defined the return value as 3-element tuple format because I did not need to update ``text`` at the time :smile: In other words, there is no good reason the format cannot ever be expanded to cover it.

Have you considered extending the format to 4-element tuple?  I think it would only require:

1) Update  documentation in ``__init__``
2a) Add ``text`` parameter to ``marshal_to`` in ``__call__`` in ``Getter`` and ``Delter``.
2b) Wrap calls in ``try...except TypeError:``, so if 4-parameter ``marshal_to`` call raises, then fall back to 3-parameter call.
3 Add conditional to ``__call__`` in ``Setter`` to support updating of ``text`` attribute if tuple length is 4 instead of 3.
4) I don't remember if there are unittests on this or not, but even so, adding one to exercise the 4th tuple element shouldn't be too hard.

Then no low-level API changes are needed, forward-compatibility is maintained, and all old code will continue to work (b/c both 3 and 4 element/parameters are accepted).

rbian:
Thanks @cevich , I will update it at your suggestion.

rbian:
@cevich , could you please review the code again?
For item 3: Add conditional to __call__ in Setter to support updating of text attribute if tuple length is 4 instead of 3.
Tuple length should either be 3 or 2, right?

cevich:
> For item 3: Add conditional to call in Setter to support updating of text attribute if tuple length is 4 instead of 3. Tuple length should either be 3 or 2, right?

Hmmm, could be, maybe I misread....Ahh yep, you're correct, it's my mistake.

> @cevich , could you please review the code again?

Changes looks really good (I have no idea if they actually work), documentation reads good, and I'm happy to see unittest updates.  I think 36747bd5fff578e5302d631c413a0e5f12c1f648 can be dropped from PR now, no?

@yumingfei 

It's been a **LONG** time since I've seen this code, I don't remember what tests/modules use it, could you look over my shoulder and double-check it's okay?  Otherwise I think it LGTM.  I'm happy to help.

yumingfei:
@cevich, I have checked the codes, and almost all codes need SubElement without text option.
So if latest code is compatible to older format, they will work fine.

yumingfei:
Applied, thanks @rbian and @cevich !

PR #1907, opened 2014-12-03
Parse interface details from domif-list output.

rbian:


virttest/utils_test/libvirt.py
```diff
@@ -1217,6 +1277,121 @@ def file_exists():
     return diskxml.xml
 
 
+def create_net_xml(net_name, params):
+    """
+    Create a new network or update an existed network xml
+    """
```
rbian: You are right, should check if the network exist or not here.

yumingfei:
Thanks~applied.

PR #1910, opened 2014-12-05
Add error info when s4 is not available

cheliu:
s4 check could be failed because the rest of guest disk space is less than guest memory.
Will print rest of disk space and gest memory and free memory in the error info

Signed-off-by: Chengyou Liu <cheliu@redhat.com>


ypu:
As the author is not working on this anymore will close it. And arrange this bug fix later.

PR #1922, opened 2014-12-16
Suffix

FengYang:
After this patch, A_min set minimum for A, A_max set maximum for A,
A_fixed set fixed value for A, which means will not covered by setting after it.

This feature is useful when we set a parameter and do not want it covered by
new value late.

We need it when running case in loops.


xutian:
LGTM,

Acked-by:  @xutian

PR #1933, opened 2014-12-24
shared.unattended: fix ubuntu timeout lost issue

xutian:
In system_set_bootable test, recordfail set to 1 when reset guest before guest boot into OS, and make_timeout function in /etc/grub.d/00_header will set timeout to -1 if GRUB_RECORDFAIL_TIMEOUT not set. And grub wait  user to select boot OS. Now set GRUB_RECORDFAIL_TIMEOUT and update grub before finished install to fix this issue.


ypu:
Thanks @xutian and @yaxue will merge it.

PR #1939, opened 2014-12-29
    utils_test.libvirt: Update prepare pool method

will-Do:
1. Use kwargs instead of optional parameters
    2. Support 'source-format' when prepare pool
    3. Support both ipv4 and ipv6 for iscsi pool
    4. Support define persistent pool
    5. Update SCSI pool XML create method

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -41,6 +41,7 @@ from virttest.libvirt_xml import vm_xml
 from virttest.libvirt_xml import xcepts
 from virttest.libvirt_xml import NetworkXML
 from virttest.libvirt_xml import IPXML
+from virttest.libvirt_xml import pool_xml
 from virttest.libvirt_xml.devices import disk
 from virttest.libvirt_xml.devices import hostdev
 from virttest.libvirt_xml.devices import controller
@@ -780,11 +781,13 @@ class PoolVolumeTest(object):
         self.selinux_bak = ""
 
     def cleanup_pool(self, pool_name, pool_type, pool_target, emulated_image,
-                     source_name=None):
+                     **kwargs):
         """
         Delete vols, destroy the created pool and restore the env
         """
         sp = libvirt_storage.StoragePool()
+        source_format = kwargs.get('source_format')
+        source_name = kwargs.get('source_name')
         try:
             if sp.pool_exists(pool_name):
                 pv = libvirt_storage.PoolVolume(pool_name)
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -828,43 +828,44 @@ def cleanup_pool(self, pool_name, pool_type, pool_target, emulated_image,
...
         :param pool_name: created pool name
         :param pool_type: dir, disk, logical, fs, netfs or else
         :param pool_target: target of storage pool
         :param emulated_image: use an image file to simulate a scsi disk
-                               it could be used for disk, logical pool
-        :param image_size: the size for emulated image
-        :param pre_disk_vol: a list include partition size to be created
-                             no more than 4 partition because msdos label
+                               it could be used for disk, logical pool, etc
+        :param kwargs: key words for specific pool
```
yumingfei: Since you replace those options with kwargs, I think you should still comment all the keys for kwargs.

virttest/utils_test/libvirt.py
```diff
@@ -915,71 +917,70 @@ def pre_pool(self, pool_name, pool_type, pool_target, emulated_image,
...
 
+        # Create/define pool
+        if pool_type == "scsi":
+            re_v = func(scsi_xml_file, debug=True)
+        else:
+            re_v = func(pool_name, pool_type, pool_target, extra, debug=True)
         if not re_v:
-            raise error.TestFail("Create pool failed.")
-
-        ret = virsh.pool_dumpxml(pool_name)
```
yumingfei: Why do not check pool and output the pool XML?

yumingfei:
Anyway, the enhancements are good to me.:)

virttest/utils_test/libvirt.py
will-Do: It's necessary to call  self.cleanup_pool() if the step failed, so we don't want to raise other exception in 'check_actived_pool' .

yumingfei: Can you deal with it to avoid raising exception?I believe the checking is useful here.

will-Do: I'm not sure why the checking is necessary here. ```check_actived_pool``` just check if the given pool is running(pool in inactive status also raise TestFail), so it's not suit for the ```pre_pool``` func any more, or we can update that func to check both active or inactive pool, but it's not what I want to do here.
And what I though the calling logical here is, cases call ```pre_pool``` to define/create a pool, if ```re_v == 1```, then we clean the env and raise TestFail, if ```re_v != 1```, then just let the cases deal with the next step result.

yumingfei: Because we may meet many problems during test and many of them are not predictable.
So debugging some information during test will be helpful for testers when case is 
failure or something wrong. 

The exit status is not readable in logs.That's why I often add debugging
information such as "XML output".
With checking pool here, we can now what has happened during test.

will-Do: Okay, I will print the pool XML here, but as i said, I don't want to touch 'check_actived_pool()' here as it only pass for active pool.

will-Do:
Thanks, just update it, and fix another problem: 
when define/create pool, it will calling 4 different virsh functions, and 'virsh.pool_create_as' return boolean, but other 3 functions return CmdResult object.

will-Do:
Thanks @yumingfei , now merge this PR.

PR #1946, opened 2015-01-06
virttest.libvirt_vm: Generate random serial log file name

Hao-Liu:
Using a deterministic log file name will cause overwrite of serial
log and only the last serial log file will be saved if multiple
log session has been established in one test.

This patch add a random suffix to prevent overwrite.


yumingfei:
Acked by: Yu Mingfei <yumingfei@cn.fujitsu.com>

will-Do:
ACK

PR #1947, opened 2015-01-07
Introduce utils_gdb module and LibvirtdSession class

Hao-Liu:
In this PR, a module utils_gdb for manipulating an inferior process in gdb is introduced.
Another commit in the PR brings a LibvirtdSession class which creates an interactive 
libvirt daemon session and can optionally use utils_gdb module debug the process.

These changes will be helpful to break at specific function call or get the backtrace info
etc.

### DIFF
--- a/virttest/utils_libvirtd.py
+++ b/virttest/utils_libvirtd.py
@@ -1,11 +1,15 @@
 """
 Module to control libvirtd service.
 """
+import re
 import logging
+import aexpect
 
 from virttest import remote, utils_misc
 from autotest.client import utils, os_dep
+from autotest.client.shared import error
 from virttest.staging import service
+from virttest.utils_gdb import GDB
 
 
 try:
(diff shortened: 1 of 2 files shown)


virttest/utils_libvirtd.py
```diff
@@ -77,6 +81,233 @@ def is_running(self):
...
+    def _start_callback(self, gdb, info, params):
+        """
+        Stop handler function triggered when gdb libvirtd started.
+
+        :param gdb: Instance of the gdb session
+        :param status: Return code of exited libvirtd session
+        """
+        self.running = True
+        params['stop-info'] = None
+
```
will-Do: Seems there are unused parameters in these private functions, do you want to leave it for feature expand or ?

will-Do:
Well done, really a useful module for debugging.

virttest/utils_libvirtd.py
Hao-Liu: No, those unused parameters are enforced since the parameters must be match what a callback function requires. We should leave them even if not used, or parameter check will fail when calling back.

yumingfei:
Very nice module! Thoughtful and useful.

virttest/utils_libvirtd.py
will-Do: Okay, thanks.

yumingfei:
Applied, thanks @Hao-Liu

PR #1929, opened 2014-12-23
utils_libguestfs: Unset color environment for guestfish test.

yumingfei:
For some Linux release(RHEL7.1), characters will have color
when we enter guestfish shell session. It may cause failure
of testcases. So covering it during testing.

Signed-off-by: Yu Mingfei <yumingfei@cn.fujitsu.com>

PR #1949, opened 2015-01-07
Disable colorful prompts for guestfish

shi2wei3:
colorful prompts could affect the output content we get


yumingfei:
Hey, @shi2wei3 , I have submitted a patch about this: #1929 ,
Could you take a look at that?

shi2wei3:
Checked, thanks for reminding , i have merged your patch.

PR #1953, opened 2015-01-08
virt: Run type script in a sequence defined by type parameter.

FengYang:
ID: 1010789

### DIFF
--- a/virt.py
+++ b/virt.py
@@ -147,7 +147,8 @@ class virt(test.test):
                         env.save()
 
                     # Run the test function
-                    for t_type, test_module in test_modules.items():
+                    for t_type in t_types:
+                        test_module = test_modules[t_type]
                         run_func = utils_misc.get_test_entrypoint_func(
                             t_type, test_module)
                         try:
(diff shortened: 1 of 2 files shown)


ypu:
@FengYang Hi should we also update the run_once in standalone_test.py?

FengYang:
On 01/08/2015 06:51 PM, Yiqiao Pu wrote:
>
> @FengYang <https://github.com/FengYang> Hi should we also update the
> run_once in standalone_test.py?
>
Maybe, will check it. Just never use it

FengYang:
Update it.

Hao-Liu:
@FengYang, Just a little curious. What's the difference after this change? Shouldn't the result be the same?

FengYang:
On 01/20/2015 03:00 PM, Hao Liu wrote:
>
> @FengYang <https://github.com/FengYang>, Just a little curious. What's 
> the difference after this change? Shouldn't the result be the same?
>
In current code. If we set type = "boot shutdown". It may run shutdown 
first, but we may want to make sure boot run firstly.

Hao-Liu:
Right, understood and Acked.

ypu:
LGTM

ypu:
Thanks @FengYang and @Hao-Liu will merge it.

PR #1954, opened 2015-01-09
virttest/utils_libguestfs.py

zcyes:
add fs_attr_ops

### DIFF
--- a/virttest/utils_libguestfs.py
+++ b/virttest/utils_libguestfs.py
@@ -2246,6 +2246,387 @@ class GuestfishPersistent(Guestfish):
         daemon) that runs inside the hypervisor.
         """
         return self.inner_cmd("debug %s %s" % (subcmd, extraargs))
+    
+    def set_e2uuid(self, device, uuid):
+        """
+        set-e2uuid - set the ext2/3/4 filesystem UUID
+
+        This sets the ext2/3/4 filesystem UUID of the filesystem on "device" to
+        "uuid". The format of the UUID and alternatives such as "clear",
+        "random" and "time" are described in the tune2fs(8) manpage.
+        """
+        return self.inner_cmd("set_e2uuid %s %s" % (device, uuid))
+
+    def get_e2uuid(self, device):
+        """
+        get-e2uuid - get the ext2/3/4 filesystem UUID
+
+        This returns the ext2/3/4 filesystem UUID of the filesystem on "device".
+        """
+        return self.inner_cmd("get_e2uuid %s" % (device))
+
+    def vfs_uuid(self, mountable):
+        """
+        vfs-uuid - get the filesystem UUID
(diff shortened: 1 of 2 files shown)


yumingfei:
Some hints for the pull request:
1.A succinct but clear title.
2.Good description for patch subject and this pull request.

So as the patches in tp-libvirt. :)

zcyes:
1. add file "guestfish_fs_attr_ops.py" for Libguestfs fis_attr_ops test, and the commands are:
(set_e2uuid, get_e2uuid, vfs_uuid, findfs_uuid) (set_uuid) (set_e2label, get_e2label, vfs_label, findfs_label) (set_label) (set_e2attrs, get_e2attrs) (set_e2generation, get_e2generation) (statvfs) (tune2fs_l) (tune2fs) (vfs_type, fsck) (mkfs/mkfs_opts) (blkid) (e2fsck) (filesystem_available) (list_filesystems) (mkfifo) (sync) (mklost_and_found) (mknod) (mknod_b) (mknod_c) (ntfsresize_opts)

zcyes:
1. add test (resize2fs, resize2fs_M, resize2fs_size)
2. image_size can be set on 'M'

shi2wei3:
Look good to me, but patches must be splited into even small pieces with more clear comment later on.
One commit fix/add one bug/feature

yumingfei:
Yes, commit message should be more clear.But codes are good to me.

BTW, maybe we should take account of the scale of utils_libguestfs module now.
It's rather big now. I really didn't realize that there be so many functions added in.

shi2wei3:
My though is to split the API encapsulation related function out, take the advantage of python's multiple inheritance, add a new class GuestfishAPI hold this functions, let GuestfishTools both inherit GuestfishPersistent and GuestfishAPI. Move class GuestfishTools and GuestfishAPI from virt-test repo to tp-libvirt repo.

zcyes:
@yumingfei 
Thanks for your comments.
You say "take account of the scale of utils_libguestfs",  We consider three way to handle this.
1. remove all apis, and just use inner.cmd(). But we don't think this is a good idea.
2. move apis to another new created file in virttest repo.
3. move apis to tp_virt repo. But we find that there's no ideal path to keep the apis file, cause no testers put their apis in tp_virt repo.

Do you have any suggestion? we are very appreciate. thanks

yumingfei:
I prefer choice 2.
Just as @shi2wei3 said, we can split APIs out of utils_libguestfs module.
Maybe we can create a new subdirectory named utils_libguestfs
(I'm afraid libguestfs may be conflict with some python-binding module, maybe not :) ),
then we can split classes in original utils_libguestfs module to LibguestfsTools, GuestfishAPI in new utils_libguestfs directory and so on.

PR #1955, opened 2015-01-09
Fix bc4bfb8, run unset under the same process for guestfish

shi2wei3:
@yumingfei Sorry, i haven't check your PR carefully, unset command shouldn't execute by utils.run, this function will fork a subprocess to run your command which will have no effect on current process and subprocess we created later like guestfish.

### DIFF
--- a/virttest/utils_libguestfs.py
+++ b/virttest/utils_libguestfs.py
@@ -191,10 +191,9 @@ class Guestfish(LibguestfsBase):
         unset_cmd = ""
         for env in color_envs:
             unset_cmd += "unset %s;" % env
-        if unset_cmd:
-            utils.run(unset_cmd, ignore_status=True)
-
-        if run_mode == "remote":
+        if run_mode == "interactive" and unset_cmd:
+            guestfs_exec = unset_cmd + " " + guestfs_exec
+        elif run_mode == "remote":
             guestfs_exec += " --listen"
         else:
             if uri:


shi2wei3:
@yumingfei Which test case your are running?
I tried type_specific.io-github-autotest-libvirt.guestfs_add.normal_test.add_disk
This test case will failed, you can check the output like 
def get_root(guestfs):
    getroot_result = guestfs.inspect_os()

yumingfei:
Same case as yours, maybe my current environment disturbed it.
I'll try it again.

yumingfei:
OK, maybe it was caused by configured environment on my machine.
BTW, do you know which file contains these guestfish environments?I can't find it now...

virttest/utils_libguestfs.py
```diff
@@ -191,10 +191,9 @@ def __init__(self, disk_img=None, ro_mode=False,
         unset_cmd = ""
         for env in color_envs:
             unset_cmd += "unset %s;" % env
-        if unset_cmd:
-            utils.run(unset_cmd, ignore_status=True)
-
-        if run_mode == "remote":
+        if run_mode == "interactive" and unset_cmd:
+            guestfs_exec = unset_cmd + " " + guestfs_exec
+        elif run_mode == "remote":
```
yumingfei: Does remote mode need to unset environments?

shi2wei3: No, remote mode won't have such problem, i have tried this.

shi2wei3: http://libguestfs.org/guestfish.1.html
ENVIRONMENT VARIABLES

Do you mean that?
There seems no src file contains a list of all these envs.

yumingfei: Thanks, that's it.
And I find there is a file on RHEL7.1 for guestfish environment: /etc/profile.d/guestfish.sh. :)

yumingfei:
LGTM now.

PR #1960, opened 2015-01-15
virttest.iscsi: Correct the regex to get all ipv6 version nodes

will-Do:
The old regex can only match node incomplete ipv6 address(eg. ::1),
after update it, it can match all ipv6 address.
And, when delete iscsi node, afte match the expected target, and delete
it, jump the loop immediately.


yumingfei:
LGTM.

will-Do:
Thanks, merge this patch.

PR #1961, opened 2015-01-15
virttest: Refactorize guest agent related functions

Hao-Liu:
Guest agent is vastly used in the tests so a single function call
to prepare guest agent is needed. Current using method set_guest_agent()
in virttest.util_test.libvirt is not so flexible. For example for tests
requiring a guest with agent channel but no agent service etc.

In this patch we implemented a method prepare_guest_agent() in libvirt_vm
class to setup guest agent as required for guest.

Major changes including:
1)  Changes get_agent_channels() and remove_agent_channels() from
    static method back to normal method since using static method
    is not straightforward. There are many misuses of static method
    in libvirt_xml module and We should change them later since
    code dependence is tangled.

2)  Besides the prepare_guest_agent() method mentioned above, two new
    methods install_package() and remove_package() are implemented
    for easily handling packages in the guest system. It only supports
    yum based package manager and others are expected to be supported
    later.

3)  The original set_guest_agent() will report a deprecate warning in
    case it is used accidentally after change.

4)  Added getenforce() and setenforce() in libvirt_xml module to
    retrieve or change guest SELinux mode for later use.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -2249,3 +2249,126 @@ class VM(virt_vm.BaseVM):
             return None
         session.close()
         return mac.strip()
+
+    def install_package(self, name):
+        """
+        Install a package on VM.
+        ToDo: Support multiple package manager.
+
+        :param name: Name of package to be installed
+        """
+        session = self.wait_for_login()
+        try:
+            # Install the package if it does not exists
+            cmd = "rpm -q %s || yum install -y %s" % (name, name)
+            status, output = session.cmd_status_output(cmd, timeout=300)
+            if status != 0:
+                raise virt_vm.VMError("Installation of package %s failed:\n%s" %
+                                      (name, output))
+        finally:
+            session.close()
+
+    def remove_package(self, name):
+        """
+        Remove a package from VM.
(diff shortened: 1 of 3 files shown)


virttest/libvirt_vm.py
```diff
@@ -2206,3 +2206,123 @@ def get_interface_mac(self, interface):
...
+            session.close()
+
+    def setenforce(self, mode):
+        """
+        Set SELinux mode in the VM.
+
+        :param mode: SELinux mode [Enforcing|Permissive|1|0]
+        """
+        self.install_package('libselinux-utils')
+        if int(mode) == 1:
```
will-Do: As you mentioned above, if 'mode' is String, here will raise ValueError.

will-Do:
Other parts LGTM.

virttest/libvirt_vm.py
Hao-Liu: Yes, you are right. I'll fix this.

Hao-Liu:
@will-Do, Fixed, thanks for the note.

will-Do:
Thx, pushed

PR #1962, opened 2015-01-15
virttest.virsh: Add domfsfreeze, domfsthaw and domtime

Hao-Liu:


will-Do:
ACK and merge.

PR #1964, opened 2015-01-15
virttest.utils_test.libvirt: Add helper function to check command result

Hao-Liu:
This helper function is an enhanced version to existing function
check_exit_status().

The original one only tests whether the exit status matches expectation.
The newer one also consider the content of the error messages if the
command is expected be fail.

Add expected fail patterns (regex) as a list for expected_fails
argument.  If test stderr matches non of these, an TestFail will emit.
If this list is empty. Command is expected to be successful. If the
command is expected be fail but the error pattern does not matter,
setting any_error to True will suffice.

If any of the skip_if list of regex matches command stderr. The test
will skip.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1129,6 +1129,48 @@ class MigrationTest(object):
         vm.connect_uri = srcuri
 
 
+def check_result(result, expected_fails=[], skip_if=[], any_error=False):
+    """
+    Check the result of a command and check command error message against
+    expectation.
+
+    :param result: Command result instance.
+    :param expected_fails: list of regex of expected stderr patterns. The check
+                           will pass if any of these patterns matches.
+    :param skip_if: list of regex of expected patterns. The check will raise a
+                    TestNAError if any of these patterns matches.
+    :param any_error: Whether expect on any error message. Setting to True will
+                      will override expected_fails
+    """
+    logging.debug("Command result:\n%s" % result)
+    if skip_if:
+        for patt in skip_if:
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1129,6 +1129,48 @@ def cleanup_dest_vm(self, vm, srcuri, desturi):
...
+    Check the result of a command and check command error message against
+    expectation.
+
+    :param result: Command result instance.
+    :param expected_fails: list of regex of expected stderr patterns. The check
+                           will pass if any of these patterns matches.
+    :param skip_if: list of regex of expected patterns. The check will raise a
+                    TestNAError if any of these patterns matches.
+    :param any_error: Whether expect on any error message. Setting to True will
+                      will override expected_fails
```
will-Do: Seems these 3 optional params are not clear literally, my suggestion: expected_error, skip_error and maybe 'any_error' can remove. If both expected_error and skip_error are [], then that means any_error is True. What do you think?

Hao-Liu: I think changing as what you said will make it impossible to check for a successful result.

will-Do: Okay

yumingfei:
So, why not remove the check_exit_status?

Hao-Liu:
There will be almost 200 implements. I just thought the impact caused by the changes might be too large. Shall we take the risk and check them all? @yumingfei

yumingfei:
It should be replaced for long-term perspective.
But we can apply this one right now, and you can go on replacing job step by step.

PR #1965, opened 2015-01-15
utils_test.__init__.py: make class VMStress support iozone tool

PandaWei:
Although the docstring of class VMStress metioned iozone,
in fact, it doesn't support increase workload using iozone tool.
This patch make it into practice.


yumingfei:
Good.

PR #1967, opened 2015-01-16
virttest.virsh: bugfix for incorrect parameter location

PandaWei:
The domain name should be specified behind the option "--domain".


PandaWei:
The old code will make wrong parameter order, just for an example
* Command: 
    /usr/bin/virsh detach-device --domain --live rhel7.0 /tmp/virt-
    test/tmp/hotplug_serial_load/xml_file1
Exit status: 1
Duration: 0.0118579864502

stderr:
error: unexpected data '/tmp/virt-test/tmp/hotplug_serial_load/xml_file1'

yumingfei:
domain_opt is used for value of "--domain", so please pass "--live" to flag_args for attach_device(...).
So NACK.

PR #1969, opened 2015-01-16
utils_misc: Add timeout parameter to yum_install.

rbian:
As it may take a long time to install packages, need a parameter to
specify timeout value.


yumingfei:
ACK

PR #1971, opened 2015-01-19
Add 16 augeas related functions. test passed,  no other effect

alex8866:


shi2wei3:
Just simple API skeleton definition, acked.

PR #1973, opened 2015-01-20
virt.virttest.qemu_vm: reboot with passed nic when session is None

xutian:
Fix nic_index miss when reboot with session=None.

ID:1183987


ypu:
LGTM

PR #1976, opened 2015-01-21
Bug fix for add-drive-opts readonly

shi2wei3:
readonly = params.get("gf_add_readonly", "no")

if readonly:

These will cause readonly never set to NULL

### DIFF
--- a/virttest/utils_libguestfs.py
+++ b/virttest/utils_libguestfs.py
@@ -501,7 +501,7 @@ class GuestfishPersistent(Guestfish):
         """
         cmd = "add-drive-opts %s" % filename
 
-        if readonly:
+        if readonly and readonly != 'no':
             cmd += " readonly:true"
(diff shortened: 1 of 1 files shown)


yumingfei:
Hmm, I think it's problem of test module.
We should use it in this mode:
readonly = "yes" == params.get(xxx).

ypu:
Agree with @yumingfei this variable should be a boolean but not a string. But maybe we are leak of the docstring in this function to show the usage of it.

shi2wei3:
Agree

yumingfei:
If it troubles you too much, I think changing to "if readonly is True" is also acceptable.

shi2wei3:
@yumingfei It's very kind of you, leave readonly to boolean type is OK to me, i'll submit a patch for tp-libvirt cases right now.

PR #1978, opened 2015-01-21
Fill missing XML elements for console and serial devices

Hao-Liu:
This patch add 'protocol_type', 'target_port', 'target_type' and 'sources' elements to console 
device XML and 'target_port', 'target_type' and 'sources' elements to console device XML.


yumingfei:
LGTM.

PR #1979, opened 2015-01-21
Fix typo error and add a new helper function remove_all_device_by_type

Hao-Liu:
Please refer to commit messages for the reasoning.

### DIFF
--- a/virttest/libvirt_xml/accessors.py
+++ b/virttest/libvirt_xml/accessors.py
@@ -507,7 +507,7 @@ class XMLAttribute(AccessorGeneratorBase):
                                              self.tag_name, create=False)
             value = element.get(self.attribute, None)
             if value is None:
-                raise xcepts.LibvirtXMLNotFoundError("Attribute %s not found"
+                raise xcepts.LibvirtXMLNotFoundError("Attribute %s not found "
                                                      "on element %s"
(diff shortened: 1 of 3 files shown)


Hao-Liu:
@yumingfei, Thank you for pointing out the mistakes. Now fixed.

Hao-Liu:
@yumingfei, The comment about "Last login:" has been flushed when I change the PR.
Quoted:
>Have you tested it? It seems that I have got opposite result...
```
>>> import re 
>>> login_str = "Last Login: " 
>>> regex1 = r"(?<![Ll]ast).*[Ll]ogin:\s*$" 
>>> regex2 = r"(?<![Ll]ast )[Ll]ogin:\s*$" 
>>> print re.match(regex1, login_str) 
<_sre.SRE_Match object at 0xa03cc8> 
>>> print re.match(regex2, login_str) 
None 
```
Answer:
The regex used here is actually to filter message like `Last login: Wed Jan 21 14:52:47 2015 from *` from expected matching string `localhost login:`. So any string starts with `Last login:` should not be matched mistakenly. Your result is fine. It's my fault for include vague commit message. Now changing. Thanks for the help.

yumingfei:
Got it, Acked.

PR #1980, opened 2015-01-21
virttest: Fix the wrong indent in handle_prompts

ypu:
The last of the function should be moved outside of the while loop.
Otherwise it will never be executed.


Hao-Liu:
Right. Never used the return value before. Acked.

yumingfei:
OK.

PR #1982, opened 2015-01-22
libvirt_xml.nwfilter_xml: Use XMLElementList for filterref

waynesun09:
Multiple filterref could be in one filter, so update filterref use
XMLElementList and marshal functions to support this.


yumingfei:
Yes, it should be.
Acked.

yumingfei:
Thanks @waynesun09

PR #1987, opened 2015-01-28
virttest: Update check_guests_proc_scsi in qemu_qtree

ypu:
Current output of qtree will include hex numbers and this makes
the transfer from string to int failed. So remove it to make the function
works as expect.


ypu:
ID: 1104911

spcui:
LGTM
Acked-by: @spcui

xutian:
LGTM,

Acked-by: @xutian

ypu:
Thanks @spcui and @xutian will merge it.

PR #1990, opened 2015-01-29
utils_test.libvirt: Support tester-defined device

yumingfei:
Currently tgtadm may be not work on RHEL7,
because they are using an interactive tool named targetcli.
So it's necessary to sidestep it when tgtadm does not work
and iscsi moduld does not support it.

Signed-off-by: Yu Mingfei <yumingfei@cn.fujitsu.com>


will-Do:
LGTM, acked.

yumingfei:
Thanks, merged.

PR #1993, opened 2015-02-03
virttest.virt_vm: Increase REBOOT_TIMEOUT value

spcui:
ID: 1088129


CongSmile:
Hi scui,

We can't just enlarge reboot_timeout without any judgement.
Or we will miss real bug.

Thanks.

ypu:
@spcui we should not update the default reboot time out here, but need update the reboot timeout for special test cases. So Nack this one too. And will close it. Thanks @CongSmile for reviewing.

PR #1995, opened 2015-02-03
virttest: Update the method of guess_type for QtreeDev

ypu:
As the newest version of qemu-kvm updated the info qtree output for
virtio block device. So we need update the function to fit it.


ypu:
ID: 1185695

xutian:
LGTM,

FengYang:
Looks good to me.

Acked-by: Feng Yang

PR #1996, opened 2015-02-04
Bug fix: libvirt_vm.py in add_serial() method

igord-daynix:
add_serial() method returned None in case VM lacks --serial option, it caused concatenation error in make_create_command() method.


yumingfei:
Yes, it should be.
ACK

will-Do:
Thanks and merge this patch.

PR #1997, opened 2015-02-05
virttest.utils_test.libvirt: Update create_net_xml function.

rbian:
Add more sections in xml to create_net_xml; 
add a parameter define_error for negative test.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -18,6 +18,7 @@ More specifically:
 
 import re
 import os
+import ast
 import logging
 import shutil
 import threading
@@ -1484,11 +1485,11 @@ def create_net_xml(net_name, params):
         if net_dns_forward:
             dns_dict["dns_forward"] = net_dns_forward
         if net_dns_txt:
-            dns_dict["txt"] = eval(net_dns_txt)
+            dns_dict["txt"] = ast.literal_eval(net_dns_txt)
         if net_dns_srv:
-            dns_dict["srv"] = eval(net_dns_srv)
+            dns_dict["srv"] = ast.literal_eval(net_dns_srv)
         if net_dns_forwarders:
-            dns_dict["forwarders"] = [eval(x) for x in
+            dns_dict["forwarders"] = [ast.literal_eval(x) for x in
                                       net_dns_forwarders]
         if net_dns_hostip:
             host_dict["host_ip"] = net_dns_hostip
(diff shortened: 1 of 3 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1436,7 +1436,7 @@ def file_exists():
     return diskxml.xml
 
 
-def create_net_xml(net_name, params):
+def create_net_xml(net_name, params, define_error=False):
```
will-Do: This should be a general func for test case, so deal with the result in your test case is more appropriate.

rbian: Ok, I will update it. Thanks

virttest/utils_test/libvirt.py
```diff
@@ -1542,9 +1574,35 @@ def create_net_xml(net_name, params):
...
+                if len(pg_bandwidth_inbound) > i:
+                    pgxml.bandwidth_inbound = eval(pg_bandwidth_inbound[i])
+                if len(pg_bandwidth_outbound) > i:
+                    pgxml.bandwidth_outbound = eval(pg_bandwidth_outbound[i])
+                if len(pg_vlan) > i:
+                    pgxml.vlan_tag = eval(pg_vlan[i])
+                netxml.set_portgroup(pgxml)
         logging.debug("New network xml file: %s", netxml)
         netxml.xmltreefile.write()
-        netxml.sync()
```
yumingfei: Do you mean that it should be done by user?

yumingfei:
Other looks fine.

virttest/utils_test/libvirt.py
rbian: Yes, user will define the network and deal with exceptions.

will-Do:
ACK
BTW, it's better to squash your commits

virttest/utils_test/libvirt.py
```diff
@@ -1501,7 +1505,33 @@ def create_net_xml(net_name, params):
...
         forward = eval(net_forward)
         if forward:
+            net_ifs = utils_net.get_net_if(state="UP")
+            # Check forward device is valid or not,
+            # if it's not in host interface list, try to set
+            # forward device to first active interface of host
+            if (forward.has_key('mode') and forward['mode'] in
+                ['passthrough', 'private', 'bridge', 'macvtap'] and
+                forward.has_key('dev') and
+                    forward['dev'] not in net_ifs):
```
Hao-Liu: I think there is a logic error. forward device will be set if `forward` doesn't have key `mode` or `dev`, or `mode` is not valid.

virttest/utils_test/libvirt.py
```diff
@@ -1501,7 +1505,33 @@ def create_net_xml(net_name, params):
...
+            net_ifs = utils_net.get_net_if(state="UP")
+            # The guest will use first interface of the list,
+            # check if it's valid or not
+            # if it's not in host interface list, try to set
+            # forward interface to first active interface of host
+            if interface[0]['dev'] not in net_ifs:
+                logging.warn("Forward interface %s is not a "
+                             " interface of host, reset to %s",
+                             interface[0]['dev'], net_ifs[0])
+            interface[0]['dev'] = net_ifs[0]
```
Hao-Liu: interface will alway be reset to `net_ifs[0]` here.

rbian: Good catch.

Hao-Liu:
@rbian. I noticed there are multiple use of `eval` in this function, which mean someone can possibly run any code by just submit a malicious config file. I think we should replace them with `ast.literal_eval` or use `json` module for safety's sake.

virttest/utils_test/libvirt.py
rbian: Yes, this is expected result. If user set a invalid mode or dev, user will deal with the error.

rbian:
That's a good advice. But I don't think user will hack test script, that's meaningless.

virttest/utils_test/libvirt.py
Hao-Liu: I see. So we just don't do anything if user intend to create an invalid XML. But what if user want to test against a non-existing interface with a correct mode? Shouldn't we let user decide whether they want to assign an interface automatically or not?

Hao-Liu:
Yes, that's really a ridiculous assumption. But we should be prepared. right? :smiling_imp:

virttest/utils_test/libvirt.py
rbian: Ok, maybe we need to add a parameter for this.

rbian:
Thanks @Hao-Liu ' help. Update the code.

Hao-Liu:
Merging this. Thanks @rbian .

PR #1998, opened 2015-02-05
utils_test.libvirt: Add create nwfilter xml function

waynesun09:
Create network filter xml use params from cfg file.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -43,6 +43,7 @@ from virttest.libvirt_xml import xcepts
 from virttest.libvirt_xml import NetworkXML
 from virttest.libvirt_xml import IPXML
 from virttest.libvirt_xml import pool_xml
+from virttest.libvirt_xml import nwfilter_xml
 from virttest.libvirt_xml.devices import disk
 from virttest.libvirt_xml.devices import hostdev
 from virttest.libvirt_xml.devices import controller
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1550,6 +1551,123 @@ def create_net_xml(net_name, params):
...
+                protocol[i] = 'None'
+            else:
+                protocol[i] = rule_dict[i]['protocol']
+                rule_dict[i].pop('protocol')
+
+                if protocol[i] in PROTOCOL_TYPES:
+                    # replace '-' with '_' in ipv6 types as '-' is not
+                    # supposed to be in class name
+                    if '-' in protocol[i]:
+                        protocol[i] = protocol[i].replace('-', '_')
```
yumingfei: Why can't we use "_" instead of "-" in parameter rule directly?

waynesun09: @yumingfei because '-' is not allowed in module name, for get the protocol class, load_xml_module method was used which find the protocol module through search protocol file name under the path. 
And dash/hyphen character '-' is not in naming style:
https://www.python.org/dev/peps/pep-0008/#id29

For why it can't be, '-' is also subtraction operator and might cause problem, more detail check:
http://stackoverflow.com/questions/2064329/why-python-does-not-allow-hyphens

yumingfei: OK, got it.

PR #2001, opened 2015-02-06
Add a parameter to bootstrap.

rbian:
Modify bootstrap parameter for other script to use "force_update" to update config files in noninteractive mode.

### DIFF
--- a/virttest/bootstrap.py
+++ b/virttest/bootstrap.py
@@ -720,7 +720,7 @@ def bootstrap(test_name, test_dir, base_dir, default_userspace_paths,
               check_modules, online_docs_url, restore_image=False,
               download_image=True, interactive=True, selinux=False,
               verbose=False, update_providers=False,
-              guest_os=defaults.DEFAULT_GUEST_OS):
+              guest_os=defaults.DEFAULT_GUEST_OS, force_update=False):
     """
     Common virt test assistant module.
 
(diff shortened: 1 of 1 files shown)


yumingfei:
But shouldn't you provide an option in script 'run'?

rbian:
There is already an option "--update-config" in run. Thanks @yumingfei

yumingfei:
I mean the place we call bootstrap(...) and this '--update-config' does not work for it.
```
line 871:
                download_image = False
                keep_image = False

            test_dir = data_dir.get_backend_dir(self.options.type)
            bootstrap.bootstrap(test_name=self.options.type, test_dir=test_dir,
                                base_dir=data_dir.get_data_dir(),
                                default_userspace_paths=default_userspace_paths,
                                check_modules=check_modules,
                                online_docs_url=online_docs_url,
                                interactive=interactive,
                                download_image=download_image,
                                selinux=self.options.selinux_setup,
                                restore_image=keep_image,
                                verbose=self.options.verbose,
                                update_providers=self.options.update_providers,
                                guest_os=(self.options.guest_os or
                                          self.option_parser.default_guest_os))
            sys.exit(0)
```

So the force_update will be always False for function bootstrap(...) as its default value.
Do I misunderstand?

rbian:
You are right! I will update it.

rbian:
@yumingfei ,
Look at the code again, "bootstrap" option in "run" should always be excuted in interactive mode, so no need to specify "force_update" in this function.
 841         if self.options.bootstrap:
...
 845             interactive = True

will-Do:
Yes, it is, ack this patch

waynesun09:
ACK

Hao-Liu:
Merge this PR. Thanks @rbian, @yumingfei, @will-Do, @waynesun09

PR #2007, opened 2015-02-11
virttest.utils_net: support blocking specific host with certain time

chuanchang:
As subject.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -1781,7 +1781,7 @@ class IPv6Manager(propcan.PropCanBase):
         Check IPv6 network connectivity
         :param client_ifname: client network interface name
         :param server_ipv6: server IPv6 address
-        ::param count: sending packets counts, default is 5
+        :param count: sending packets counts, default is 5
         """
         try:
             os_dep.command("ping6")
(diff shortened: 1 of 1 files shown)


virttest/utils_net.py
```diff
@@ -3014,3 +3014,33 @@ def check_listening_port_remote_by_service(server_ip, server_user, server_pwd,
     except:
         if session:
             session.close()
+
+
+def block_specific_ip_by_time(ip_addr, block_time=None, runner=None):
```
will-Do: Good func, just a tiny suggestion about the func name, as the ```block_time``` could be omitted by caller and the method you use is by iptables, so my suggestion is ```block_ip_by_iptables```, what do you think?

chuanchang: @will-Do, the name 'block_ip_by_iptables' is too common, which isn't what I want, this funcs is used for blocking specific host IP with certain time.

lmr:
You have to fix your docstrings. Per Travis:

        2 ERRORs/WARNINGs detected while building the html docs:
        1) /home/travis/build/autotest/virt-test/virttest/utils_net.py:docstring of virttest.utils_net.block_specific_ip_by_time:4: ERROR: Unexpected indentation.
        2) /home/travis/build/autotest/virt-test/virttest/utils_net.py:docstring of virttest.utils_net.block_specific_ip_by_time:6: WARNING: Block quote ends without a blank line; unexpected unindent.

will-Do:
@waynesun09 @Hao-Liu 
As ajia is need this PR merged for the jenkins job, so please help to reveiw it

virttest/utils_net.py
```diff
@@ -3028,3 +3028,28 @@ def check_listening_port_remote_by_service(server_ip, server_user, server_pwd,
...
+    N >=1, the ${time} is [hours|minutes|seconds],etc, the default is '1 seconds'
+    :param runner: command runner, it's a remote session
+    """
+    cmd = "iptables -A INPUT -s %s -m time --kerneltz --timestart \
+           $(date +%%H:%%M:%%S) --timestop $(date --date='+%s' +%%H:%%M:%%S) \
+           -j DROP" % (ip_addr, block_time)
+    list_rules = "iptables -L"
+    try:
+        if runner is None:
+            runner = local_runner
```
PandaWei: It's more reasonable that move the judgment out of try/catch.

chuanchang: @PandaWei，older python cannot support the full exception syntax. please read 'Catching Exceptions' section in CODING_STYLE under the virt-test directory.

 The accepted formats for exception catching are:
     try...except...
     try...finally...
     try...except...else...
     try...finally...else...

PandaWei: Maybe There's some mistake
I mean that you *don't* need to check the runner is None or not  *in the try/catch block.

chuanchang: @PandaWei, it's a condition to check local 'iptables' command.

PandaWei: Nope....
You don't understand what i want to say...
I mean the check for runner would never throw any except.
you should check it before try/catch block.
Please refer to virttest/utils_net.py for more cases.

virttest/utils_net.py
```diff
@@ -3028,3 +3028,28 @@ def check_listening_port_remote_by_service(server_ip, server_user, server_pwd,
...
+            try:
+                os_dep.command("iptables")
+            except ValueError, details:
+                raise error.TestNAError(details)
```
PandaWei: and  I think it's better to check iptables exists or not at the beginning of block_specific_ip_by_time.
all of steps that before os_dep.command("iptables"),  turn meaningless if iptables doesn't exist.

lmr:

Thanks!

PR #2009, opened 2015-02-13
libvirt_xml.devices.interface: Add backend slot to interface xml.

rbian:


will-Do:
ACK and merge the tiny patch

PR #2010, opened 2015-02-15
virsh: Add new virsh commands

will-Do:
Including: domstats, cpu-models, freepages, domcapabilities, metadata,
net-dhcp-leases, qemu-monitor-event, net-event, event.


waynesun09:
ACK

Hao-Liu:
ACK and merged. Thanks @will-Do @waynesun09.

PR #2015, opened 2015-02-19
escape '?' wildcard when querying qemu devices

rbbratta:
qemu -device ? will fail if there is a single
char file in the cwd.

Triggered by running a command with 2&>1 which
generated a file named '1' in the cwd.

This caused us to call 'qemu -device 1' which
fails.


will-Do:
Yes, it is.
ACK this fix, thanks.

Hao-Liu:
ACK and merged. Thanks @rbbratta @will-Do.

PR #2016, opened 2015-02-20
utils_net.py: Fixed wrong module reference (utils -> utils_misc)

svirt:
There was a call 'utils.wait_for()' added in the commit 6e6e88de which
caused an error as the module 'utils' has no attributes 'wait_for()'.
The function 'wait_for()' is implemented in the module 'utils_misc'.


PandaWei:
Hi, 
Maybe you have a mistake.
Indeed, wait_for is not  implemented in the module utils.
but its' implement source code had included in client/utils.py.
please refer to the head "from autotest.client.shared.utils import *".

and there's a piece of advice.
you should let your commit pass the check of CI before pushing.

svirt:
Hello,
Thank you for the information.
I figured out that the problem was in autotest-framework version, I upgraded it to 0.16.1 and the test passed without any modification. So there is no need to have my this modification (and pull request).

For CI failure, I watched the log and seems it's a general pep8 failure which was not related to the particular python file modification

PR #2020, opened 2015-02-25
utils_test.libvirt: add function to hotplug/unhotplug cpu

PandaWei:
the hotplug/unhotplug operation can be supported by
setvcpus and qemu-monitor-command.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -2369,3 +2369,43 @@ def update_vm_disk_source(vm_name, disk_source_path, source_type="file"):
     logging.debug("The new VM XML:\n%s", vmxml.xmltreefile)
     vmxml.sync()
     return True
+
+
+def hotplug_domain_vcpu(domain, count, by_virsh=True, hotplug=True):
+    """
+    Hot-plug/Hot-unplug vcpu for domian
+
+    :param domain:   Domain name, id, uuid
+    :param count:    to setvcpus it's the current vcpus number,
+                     but to qemu-monitor-command,
+                     we need to designate a specific CPU ID.
+                     The default will be got by (count - 1)
+    :param by_virsh: True means hotplug/unplug by command setvcpus,
+                     otherwise, using qemu_monitor
+    :param hotplug:  True means hot-plug, False means hot-unplug
+    """
+    if by_virsh:
+        result = virsh.setvcpus(domain, count, "--live", debug=True)
+    else:
+        if hotplug:
+            cpu_opt = "cpu-add"
+        else:
+            cpu_opt = "cpu-del"
+            # Note: cpu-del is supported currently, it will return error.
+            # as follow,
+            # {
+            #    "id": "libvirt-23",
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -2369,3 +2369,33 @@ def update_vm_disk_source(vm_name, disk_source_path, source_type="file"):
     logging.debug("The new VM XML:\n%s", vmxml.xmltreefile)
     vmxml.sync()
     return True
+
+
+def hotplug_domain_vcpu(domain, count, by_virsh=True, hotplug=True):
+    """
+    Hotplug/Unhotplug vcpu for domian
```
will-Do: Should be "Hot-plug or hot-unplug"

virttest/utils_test/libvirt.py
```diff
@@ -2369,3 +2369,33 @@ def update_vm_disk_source(vm_name, disk_source_path, source_type="file"):
...
+                     otherwise, using qemu_monitor
+    :param hotplug:  True means hotplug, False means unhotplug
+    """
+    if by_virsh:
+        result = virsh.setvcpus(domain, count, "--live", debug=True)
+    else:
+        if hotplug:
+            cpu_opt = "cpu-add"
+        else:
+            cpu_opt = "cpu-del"
```
will-Do: #vim tp-libvirt/libvirt/tests/src/virsh_cmd/domain/virsh_setvcpus.py +247:
```
    ┊   ┊   # setvcpu/hotplug is only available as of qemu 1.5 and it's still
    ┊   ┊   # evolving. In general the addition of vcpu's may use the QMP
    ┊   ┊   # "cpu_set" (qemu 1.5) or "cpu-add" (qemu 1.6 and later) commands.
    ┊   ┊   # The removal of vcpu's may work in qemu 1.5 due to how cpu_set
    ┊   ┊   # can set vcpus online or offline; however, there doesn't appear
    ┊   ┊   # to be a complementary cpu-del feature yet, so we can add, but
    ┊   ┊   # not delete in 1.6.

```

PandaWei: Yes, you are right.
[root@localhost virt-test]# virsh qemu-monitor-command 15 --pretty '{"execute":"cpu-del","arguments":{"id":7}}'
{
    "id": "libvirt-23",
    "error": {
        "class": "CommandNotFound",
        "desc": "The command cpu-del has not been found"
    }
}

virttest/utils_test/libvirt.py
PandaWei: Thanks for your reminder.

PandaWei:
Updated it.

virttest/utils_test/libvirt.py
PandaWei: I think it's better that check the result at test case.
so i add some explanation.

will-Do:
OK

liyangfnst:
ACK, merge it now.

PR #2023, opened 2015-02-26
unattended_install: Change serial log name accordingly

Hao-Liu:
Serial log file name has been changed with a random suffix in commit
b2a48d5, but the reference to this log file in unattended install has
not changed accordingly.

This patch fix this problem and remove an unnecessary check of serial
ports since this is already on in vm.create_serial_console()


will-Do:
ACK

sathnaga:
ACK, Working fine with this patch :+1:

PR #2025, opened 2015-02-26
utils_misc: Add fuction get node cpu topology info

waynesun09:
Add new get cpu topology dict function, retrieve the cpu topology
info in certain node from sysfs.

Also fix get node cpus problem, the findall regex need match to
the end of line, and set extra_cpus from node 0 when current cpus
node not as 0.

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -1378,7 +1378,7 @@ class NumaInfo(object):
         cmd = utils.run("numactl --hardware")
         try:
             node_distances = cmd.stdout.split("node distances:")[-1].strip()
-            node_distance = re.findall("%s:" % node_id, node_distances)[0]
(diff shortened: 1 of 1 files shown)


will-Do:
There's a method ```get_cpu_info``` in virttest/utils_misc.py, which contains the cpu topology info, I guess you can use it directly.

waynesun09:
@will-Do The get_cpu_topology in NumaNode is for get the detail cpu info with soket_id, core_id, siblings of certain cell in current NumaNode instance.
The get_cpu_info use 'lscpu' command to get basic host cpu info, they are not the same.

will-Do:
Yes, it's different.
ACK this PR

waynesun09:
@will-Do thx, merging now.

PR #2039, opened 2015-03-03
CentOS 7.0: Fix kickstart file for unattended install

igord-daynix:
Enable dhclient on boot for eth0 network interface.


lmr:
Thanks @igord-daynix.

PR #2042, opened 2015-03-04
libvirt_xml.vm_xml: fix typo that leads to attribute error

PandaWei:
The commit e310f272f9 introduces this problem.
The error message as following,
  AttributeError: 'VMCPUXML' object has no attribute 'mashal_to_cell'


will-Do:
Good catch, just merge this tiny fix, thanks

PR #2048, opened 2015-03-06
virttest/libvirt_vm: check the status of qemu-guest-agent

PandaWei:
add steps to check the status of qemu-guest-agent service
after starting or stopping it.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -2315,6 +2315,12 @@ class VM(virt_vm.BaseVM):
         self.install_package('qemu-guest-agent')
 
         session = self.wait_for_login()
+
+        def _is_ga_running():
+            return (not session.cmd_status("pgrep qemu-ga"))
+
+        def _is_ga_finished():
+            return (session.cmd_status("pgrep qemu-ga") == 1)
         try:
             # Start/stop qemu-guest-agent
             if start:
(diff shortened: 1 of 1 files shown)


virttest/libvirt_vm.py
```diff
@@ -2315,6 +2315,9 @@ def prepare_guest_agent(self, prepare_xml=True, channel=True, start=True):
         self.install_package('qemu-guest-agent')
 
         session = self.wait_for_login()
+
+        def _is_qa_running():
```
Hao-Liu: We usually abbreviate qemu-guest-agent as qemu-ga or simply ga. I don't think qa is a intuitive name.

virttest/libvirt_vm.py
```diff
@@ -2325,6 +2328,10 @@ def prepare_guest_agent(self, prepare_xml=True, channel=True, start=True):
             if status != 0:
                 raise virt_vm.VMError("Start/stop of qemu-guest-agent "
                                       "failed:\n%s" % output)
+            # Check qemu-guest-agent status
+            if utils_misc.wait_for(_is_qa_running, timeout=60):
```
Hao-Liu: What if stop command is issue but qemu-ga process not exit immediately? Maybe we should also wait for qemu-ga exit for a specific time as what you do to wait for start.

PandaWei: Yes, your suggestion sounds good.
I will update as it.

Hao-Liu:
Now this PR is good to me. Thanks @PandaWei.

waynesun09:
ACK

PR #2051, opened 2015-03-06
virttest.libvirt_xml: Fix misuse of remove_by_xpath

Hao-Liu:
Original implementation of remove_by_xpath() only remove the first
occurrence of found elements. This leads to some misunderstanding and
misuse of this function.

This patch adds an option remove_all to make it able to remove all
found elements but retains the default to remove just once.


waynesun09:
ACK

PandaWei:
ACK, Good PR

PR #2052, opened 2015-03-06
libvirt.py: Fixed the bug related to VG recreation failure (moved to setup_or_cleanup_iscsi())

svirt:
The VG creation failure with error 'incorrect metadata area header checksum'
occurred during multiple test runs. The solution is to call 'vgscan --cache'
in order to rebuild the LVM cache after the disk was deleted at the end of
the test

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -470,6 +470,7 @@ def setup_or_cleanup_iscsi(is_setup, is_login=True,
         _iscsi.emulated_id = _iscsi.get_target_id()
         _iscsi.cleanup()
         utils.run("rm -f %s" % emulated_path)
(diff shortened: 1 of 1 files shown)


waynesun09:
I suggested move the vg metadata code here, even though after iscsi setup not each case have vg created (e.g. logical disk will create vg and lv, block disk will not).
I see no harm here, and did not see a better place for this, the checking of whether vg created in cases will not be easy since you need check extra params to confirm it. 
@Hao-Liu @will-Do what do you think?

Hao-Liu:
Since it's not trivial to detect the very condition whether the cache should be rebuilt. I'm okay with this idea to rebuild it anyhow. Ack.

will-Do:
Yes, no harm for our test, applying.

PR #2056, opened 2015-03-09
utils_misc: Fix get_cpu_info when without session

waynesun09:
The system_output return string, while map need output to be list
of output lines.

### DIFF
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -1272,10 +1272,10 @@ def get_cpu_info(session=None):
     cpu_info = {}
     cmd = "lscpu"
     if session is None:
-        output = utils.system_output(cmd, ignore_status=True)
+        output = utils.system_output(cmd, ignore_status=True).splitlines()
     else:
         try:
-            output = session.cmd_output(cmd).strip().splitlines()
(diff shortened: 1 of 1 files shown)


Hao-Liu:
Ack

virttest/utils_misc.py
```diff
@@ -1272,7 +1272,7 @@ def get_cpu_info(session=None):
     cpu_info = {}
     cmd = "lscpu"
     if session is None:
-        output = utils.system_output(cmd, ignore_status=True)
+        output = utils.system_output(cmd, ignore_status=True).splitlines()
```
PandaWei: Is it need to call strip()?

waynesun09: @PandaWei strip is called in map(), so not necessarily.

PandaWei: Thanks for your answer,
the following statement doesn't need to call strip().
we'd better to keep consistent.
do you think so?

waynesun09: @PandaWei ok, updating.

PandaWei:
ACK

waynesun09:
@Hao-Liu @PandaWei Thanks, merging now.

PR #2057, opened 2015-03-10
virttest/utils_libguestfs: fix bug for guestfish

PandaWei:
The c51a2b2 introduces a bug,
which leads program would never enter 'else' block.
So, even if we specify option '-i',
it still report the following error message,
  libguestfs:
  error: umount_all: call launch before using this function
  (in guestfish, don't forget to use the 'run' command)


Hao-Liu:
ACK and quick merged for bug fixing.

PR #2058, opened 2015-03-10
utils_test.libvirt: Ignore vgscan failure

waynesun09:
As lvmetad could not running, the vgscan will fail, do not raise
error and fail then.


Hao-Liu:
Ack and merged.

PR #2061, opened 2015-03-11
utils_test/libguestfs: check if file exists before removing

PandaWei:
check is compulsory, otherwise,
it will report the following in same cases,
  OSError: [Errno 2] No such file or directory


Hao-Liu:
Since this fix is straightforward. Ack and Merge without testing.

PR #2062, opened 2015-03-11
virsh: Correct metadata --set option

will-Do:
1. For --set option, the value should be quoted, so unity all single
quote to double quotes.
2. Fix a wrong variable


Hao-Liu:
Ack

will-Do:
Merge this tiny fix.

PR #2067, opened 2015-03-13
virttest.utils_test.libvirt: update vm disk source functions

chuanchang:

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -2403,38 +2403,44 @@ def do_migration(vm_name, uri, extra, auth_pwd, auth_user="root",
         return False
 
 
-def update_vm_disk_source(vm_name, disk_source_path, source_type="file"):
+def update_vm_disk_source(vm_name, disk_source_path="",
+                          disk_image_name="",
+                          source_type="file"):
     """
     Update disk source path of the VM
 
     :param source_type: it may be 'dev' or 'file' type, which is default
     """
-    if not os.path.isdir(disk_source_path):
+    if not disk_source_path or not os.path.isdir(disk_source_path):
         logging.error("Require disk source path!!")
         return False
+
     # Prepare to update VM first disk source file
     vmxml = vm_xml.VMXML.new_from_dumpxml(vm_name)
     devices = vmxml.devices
     disk_index = devices.index(devices.by_device_tag('disk')[0])
     disks = devices[disk_index]
-    disk_source = disks.source.get_attrs().get(source_type)
-    logging.debug("The disk source file of the VM: %s", disk_source)
-    if not os.path.exists(disk_source):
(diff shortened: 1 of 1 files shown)


will-Do:
ack

virttest/utils_test/libvirt.py
```diff
@@ -2403,38 +2403,44 @@ def do_migration(vm_name, uri, extra, auth_pwd, auth_user="root",
...
+def update_vm_disk_source(vm_name, disk_source_path="",
+                          disk_image_name="",
+                          source_type="file"):
     """
     Update disk source path of the VM
 
     :param source_type: it may be 'dev' or 'file' type, which is default
     """
-    if not os.path.isdir(disk_source_path):
+    if not disk_source_path or not os.path.isdir(disk_source_path):
```
Hao-Liu: Why make argument `disk_source_path` optional if it is required?

virttest/utils_test/libvirt.py
```diff
@@ -2403,38 +2403,44 @@ def do_migration(vm_name, uri, extra, auth_pwd, auth_user="root",
...
-    return True
+    try:
+        disks.source = disks.new_disk_source(**{'attrs': {'%s' % source_type:
+                                                          "%s" % new_disk_source}})
+        # SYNC VM XML change
+        vmxml.devices = devices
+        logging.debug("The new VM XML:\n%s", vmxml.xmltreefile)
+        vmxml.sync()
+        return True
+    except:
```
Hao-Liu: This general catch of exception will conceal the real problem. I suggest only catch the expected errors and throw the others or just let the caller decide.

virttest/utils_test/libvirt.py
```diff
@@ -2403,38 +2403,44 @@ def do_migration(vm_name, uri, extra, auth_pwd, auth_user="root",
         return False
 
 
-def update_vm_disk_source(vm_name, disk_source_path, source_type="file"):
+def update_vm_disk_source(vm_name, disk_source_path="",
+                          disk_image_name="",
```
PandaWei: it's better that place the new parameter at the end of the parameter list.
If so, it can prevent the bad influence to the older caller

chuanchang: @PandaWei, for now, my test cases are only one caller, I have SYNCed it with this change, so don't worry about it,

virttest/utils_test/libvirt.py
chuanchang: @Hao-Liu, yeah, I see, but I don't know which certain exception is.

Hao-Liu: Then I thinks we should remove try..except or we will never know what exception happens. What do you think?

lmr:

Thanks!

PR #2068, opened 2015-03-13
utils_test: Revert a mistakenly removed import

Hao-Liu:
import of submodules are mistakenly removed as unused import in 95509f0a94aea9d60dd726a6e0de30dcfcf2f251.

And a unused import of utils_cgroup is removed.

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -36,12 +36,10 @@ from virttest import aexpect, utils_misc, virt_vm, data_dir, utils_net
 from virttest import storage, asset, bootstrap, remote
 import virttest
 
-
-try:
-    from virttest.staging import utils_cgroup
-except ImportError:
-    # TODO: Obsoleted path used prior autotest-0.15.2/virttest-2013.06.24
-    from autotest.client.shared import utils_cgroup
+# Import submodules, should not be considered as unused import
+import libvirt
+import qemu
+import libguestfs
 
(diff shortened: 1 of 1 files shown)


xutian:
Not see where use it, so why we need import these modules.

Thanks,
Xu

Hao-Liu:
Hi @xutian, I revert these lines to prevent error `AttributeError: 'module' object has no attribute 'libvirt'` when using these sub-modules after this commit merged. I'm not sure what exactly happens but introduce this imports will cure the symptom. I know it's ugly this way and I'll try to find the root cause of this.

Hao-Liu:
Nailed it!
There are many reference in our test providers import package `utils_test` as a module in following way:
```
from virttest import utils_test

utils_test.qemu.do_something()
```
This way by importing package as module will cause python only import the content of `__init__.py` in the package directory, but ignoring the submodules. Thus when we referencing our submodules, an error `AttributeError: 'module' object has no attribute 'qemu'` is emitted.

One solution to this is explicitly import these submodules in `__init__.py`, or we have to find all the references to the submodules and change them. For example from `from virttest import utils_test` to `from virttest.utils_test import qemu`.

waynesun09:
ACK, the __init__ file could be used for expose package modules.

ypu:
@Hao-Liu for the fix. This looks good for me.

Thanks @waynesun09 will merge it.

PR #2064, opened 2015-03-12
shared.cfg.guest-os: Add RHEL.7.1 guest config

spcui:
1. Add RHEL.7.1 guest config
2. Small update for RHEL.7.0 config

Signed-off-by: Shuping Cui <scui@redhat.com>

PR #1994, opened 2015-02-03
add variants grub_file in SLES and Ubuntu cfg,delete variants grub_file in RHEL7.0 cfg

PyLearner:
add variants grub_file in SLES and Ubuntu cfg,delete variants grub_file in RHEL7.0 cfg

PR #2069, opened 2015-03-16
pullreq #1994 is conflicted with #2064,solve the confilict

PyLearner:


PyLearner:
pullreq #1994 is conflicted with #2064,solve the confilict

PR #2072, opened 2015-03-16
test_setup: Add hugepage numa node functions

waynesun09:
Add function to get host supported page sizes, and functions get or
set per-node huge page size.

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -287,6 +287,8 @@ class HugePageConfig(object):
         self.deallocate = params.get("hugepages_deallocate", "yes") == "yes"
         self.hugepage_path = '/mnt/kvm_hugepage'
         self.kernel_hp_file = '/proc/sys/vm/nr_hugepages'
+        self.pool_path = "/sys/kernel/mm/hugepages"
+        self.sys_node_path = "/sys/devices/system/node"
         self.hugepage_size = self.get_hugepage_size()
         self.hugepage_force_allocate = params.get("hugepage_force_allocate",
                                                   "no")
@@ -362,6 +364,63 @@ class HugePageConfig(object):
 
         return target_hugepages
 
+    def get_multi_supported_hugepage_size(self):
+        """
+        As '/proc/meminfo' only show default huge page size, this function is
+        for get huge page size of multiple huge page pools.
+
+        For each huge page size supported by the running kernel, a
+        subdirectory will exist, of the form:
+
+            hugepages-${size}kB
+
+        under /sys/kernel/mm/hugepages, get the support size and return a list.
+
+        :return: supported size list in kB unit
+        """
+        hugepage_size = []
+        if os.path.isdir(self.pool_path):
+            for path_name in os.listdir(self.pool_path):
+                logging.debug("path name is %s" % path_name)
+                if os.path.isdir("%s/%s" % (self.pool_path, path_name)):
+                    hugepage_size.append(path_name.split('-')[1][:-2])
+                    logging.debug(path_name.split('-')[1][:-2])
+            return hugepage_size
+        else:
+            raise ValueError("Root hugepage control sysfs directory %s did not"
+                             " exist" % self.pool_path)
+
+    def get_node_num_huge_pages(self, node, pagesize):
+        """
+        Get number of pages of certain page size under given numa node.
+
+        :param node: string or int, node number
+        :param pagesize: string or int, page size in kB
+        :return: int, node huge pages number of given page size
+        """
+        node_page_path = "%s/node%s" % (self.sys_node_path, node)
+        node_page_path += "/hugepages/hugepages-%skB/nr_hugepages" % pagesize
+        if not os.path.isfile(node_page_path):
+            raise ValueError("%s page size nr_hugepages file of node %s did "
+                             "not exist" % (pagesize, node))
+        out = utils.system_output("cat %s" % node_page_path)
+        return int(out)
+
+    def set_node_num_huge_pages(self, num, node, pagesize):
+        """
+        Set number of pages of certain page size under given numa node.
+
+        :param num: string or int, number of pages
+        :param node: string or int, node number
+        :param pagesize: string or int, page size in kB
+        """
+        node_page_path = "%s/node%s" % (self.sys_node_path, node)
+        node_page_path += "/hugepages/hugepages-%skB/nr_hugepages" % pagesize
+        if not os.path.isfile(node_page_path):
+            raise ValueError("%s page size nr_hugepages file of node %s did "
+                             "not exist" % (pagesize, node))
+        utils.system("echo %s > %s" % (num, node_page_path))
+
     @error.context_aware
     def set_hugepages(self):
         """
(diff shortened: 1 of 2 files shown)


PandaWei:
Do you intend to submit new cases that call these new functions , such as  get_multi_supported_hugepage_size  and  set/get_node_num_huge_pages?
If not,  what's the usefulness?

virttest/staging/utils_memory.py
```diff
@@ -107,10 +107,74 @@ def get_transparent_hugepage():
...
+
+def get_node_num_huge_pages(node, pagesize):
+    """
+    Get number of pages of certain page size under given numa node.
+
+    :return: node huge pages number of given page size
+    """
+    node_page_path = "/sys/devices/system/node/node%s" % node
+    node_page_path += "/hugepages/hugepages-%skB/nr_hugepages" % pagesize
+    out = utils.system_output("cat %s" % node_page_path)
```
ypu: Is it better to check if the path is exist before you try to cat it?

virttest/staging/utils_memory.py
```diff
@@ -107,10 +107,74 @@ def get_transparent_hugepage():
...
+    return out
+
+
+def set_node_num_huge_pages(num, node, pagesize):
+    """
+    Set number of pages of certain page size under given numa node.
+    """
+    node_page_path = "/sys/devices/system/node/node%s" % node
+    node_page_path += "/hugepages/hugepages-%skB/nr_hugepages" % pagesize
+    utils.run("echo %s > %s" % (num, node_page_path), ignore_status=False)
```
ypu: The same above, and any special reason that we should ignore_status here?

virttest/staging/utils_memory.py
```diff
@@ -107,10 +107,74 @@ def get_transparent_hugepage():
...
 
 
+def get_num_huge_pages():
+    """
+    Get host default size huge pages number
+
+    :return: int, default size huge page number
+    """
+    result = utils.system_output('/sbin/sysctl -n vm.nr_hugepages')
+    return int(result)
```
PandaWei: There's a file with the same name at   virttest/staging/utils_memory.py 
def get_num_huge_pages():
    return read_from_meminfo('HugePages_Total')

virttest/staging/utils_memory.py
```diff
@@ -107,10 +107,74 @@ def get_transparent_hugepage():
...
+    utils.run("echo %s > %s" % (num, node_page_path), ignore_status=False)
+
+
 def set_num_huge_pages(num):
+    """
+    Set host number of huge pages of default page size
+
+    :param num: int, number of total default huge page size
+    """
     utils.system('/sbin/sysctl vm.nr_hugepages=%d' % num)
```
ypu: This already can be do in test_setup.py under HugePageConfig. Do you think about using that one for your test?

virttest/staging/utils_memory.py
waynesun09: @ypu utils.run will directly raise error here if path not exist.

virttest/staging/utils_memory.py
waynesun09: @ypu same here, raise run command error.

virttest/staging/utils_memory.py
waynesun09: @ypu the function is exist before my update here, I only update docstring to make it more clear. 

And for HugePageConfig class, it did not fit for libvirt testing, as:
- libvirt will config hugepage mount path and mount it in libvirt qemu.conf
- per-node different page size setting not supported in the class
- the only function could be used is with same purpose here

The other functions added here are common memory setting, even though related to hugepages, so consider them could be used more generally.

The utils_test module did import staging.utils_memory module to handle other memory related operation, so if needed, could update utils_test.HugePageConfig function to use new added utils_memory functions to handle more specified hugepage setting.

ypu: The __init__() function in that class will do nothing setup but only some variables. So it will not mount or setup any thing in the host. So I guess it can also work on libvirt part.

test_setup.py(not utils_test) is the module that have HugePageConfig. And yes it is not a bad idea to let it call the commen functions here and if so we may need move all common use functions from it to here. But seems for the modules in staging we need send two pullreqs one for virt-test and another for autotest.

And another problem for this function here. You use sysctl to set up it which may not fit every Linux OS.

virttest/staging/utils_memory.py
ypu: Yes but the error msg will be ugly as command exist with none zero status but not the directly reason. It will make it a little confused for people who is new hands in autotest.

waynesun09:
@PandaWei as dependency between test provider and virt-test, depend PR in virt-test could be raised before test provider PR, it's also good to stimulate conversation, like what's show here.
So after intense testing and code refactoring, the tp-libvirt PR also raised. Thanks for the concern on the PR, now could you help check both PRs. Thanks!

@ypu I decided to move the functions into HugePageConfig class without disrupt anything, hope this works for you, pls help check again. Thanks!

virttest/test_setup.py
```diff
@@ -362,6 +362,58 @@ def get_target_hugepages(self):
...
+    def set_node_num_huge_pages(self, num, node, pagesize):
+        """
+        Set number of pages of certain page size under given numa node.
+        """
+        node_page_path = "/sys/devices/system/node/node%s" % node
+        node_page_path += "/hugepages/hugepages-%skB/nr_hugepages" % pagesize
+        if not os.path.isfile(node_page_path):
+            raise ValueError("%s page size nr_hugepages file of node %s did "
+                             "not exist" % (pagesize, node))
+        utils.run("echo %s > %s" % (num, node_page_path), ignore_status=False)
```
PandaWei: There's a little suggestion...
replace utils.run with utils.system(), Just keep consistence with the othe part of class HugePageConfig.
In fact, utils.system will call utils.run.
so if you wouldn't  like to change it,  it's OK for me

Looks Good,  ACK.

waynesun09:
@PandaWei updated, thx.

virttest/test_setup.py
```diff
@@ -362,6 +362,64 @@ def get_target_hugepages(self):
...
+        For each huge page size supported by the running kernel, a
+        subdirectory will exist, of the form:
+
+            hugepages-${size}kB
+
+        under /sys/kernel/mm/hugepages, get the support size and return a list.
+
+        :return: supported size list in kB unit
+        """
+        pool_path = "/sys/kernel/mm/hugepages"
```
chuanchang: Just a suggestion, we may put 'pool_path = "/sys/kernel/mm/hugepages"' as default parameter in get_multi_supported_hugepage_size().

virttest/test_setup.py
```diff
@@ -362,6 +362,64 @@ def get_target_hugepages(self):
...
+
+    def get_node_num_huge_pages(self, node, pagesize):
+        """
+        Get number of pages of certain page size under given numa node.
+
+        :param node: string or int, node number
+        :param pagesize: string or int, page size in kB
+        :return: int, node huge pages number of given page size
+        """
+        node_page_path = "/sys/devices/system/node/node%s" % node
```
chuanchang: For common part like node_page_path = "/sys/devices/system/node/", you can put it into __init__() if you want, of course, it's not a big deal.

waynesun09:
@chuanchang I've followed your suggestion and moved both pool_path and sys node path to init, thanks.

waynesun09:
Thanks guys, merging this now.

PR #2091, opened 2015-03-23
Cleanup get IP address  releated function

xutian:
After fef5c4c5cd2da1b7509283f3b3db06166c4c8615 applied, not we need cleanup code for implement ip_pool.

thanks,
Xu

PR #2092, opened 2015-03-23
virttest.utils_net: Try default route if selected route could not work.

FengYang:
When host have multi network card in same subnet, guset may fail at
mac ip verify. Because code may ge a wrong device from ip route.

### DIFF
--- a/virttest/utils_net.py
+++ b/virttest/utils_net.py
@@ -2659,20 +2659,27 @@ def verify_ip_address_ownership(ip, macs, timeout=60.0):
 
     # Get the name of the bridge device for ip route cache
     ip_cmd = utils_misc.find_command("ip")
-    ip_cmd = "%s route get %s" % (ip_cmd, ip)
-    o = commands.getoutput(ip_cmd)
-    dev = re.findall(r"dev\s+\S+", o, re.I)
-    if not dev:
-        logging.debug("No dev in route table to %s: %s" % (ip, o))
+    ip_cmd = "%s route get %s; %s route | grep default" % (ip_cmd, ip, ip_cmd)
+    output = commands.getoutput(ip_cmd)
+    devs = re.findall(r"dev\s+\S+", output, re.I)
+    checked_devs = []
+    if not devs:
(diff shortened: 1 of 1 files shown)


FengYang:
id : 1203515

xutian:
@FengYang 

I think verify_ip_mac_owner_ship inneed in current framework, so I drop it in https://github.com/autotest/virt-test/pull/2091

thanks,
Xu

FengYang:
On 03/23/2015 04:52 PM, Xu Tian wrote:
>
> @FengYang <https://github.com/FengYang>
>
> I think verify_ip_mac_owner_ship inneed in current framework, so I
> drop it in #2091 <https://github.com/autotest/virt-test/pull/2091>
>
> thanks,
> Xu
>

I think we still need it.  Or sometimes we do not know why could not
login.  Or we may try to login a old ip which already used by another
VM, It may effect other case.  So verify the ip and mac before using it
is always a good idea. That is also why we add them before.

We just need disable it for some devices, e.g macvtap and sr-iov.

xutian:
On 03/23/2015 05:23 PM, Feng Yang wrote:
>
>
> On 03/23/2015 04:52 PM, Xu Tian wrote:
> >
> > @FengYang <https://github.com/FengYang>
> >
> > I think verify_ip_mac_owner_ship inneed in current framework, so I
> > drop it in #2091 <https://github.com/autotest/virt-test/pull/2091>
> >
> > thanks,
> > Xu
> >
>
> I think we still need it. Or sometimes we do not know why could not
> login. Or we may try to login a old ip which already used by another
> VM, It may effect other case. So verify the ip and mac before using it
> is always a good idea. That is also why we add them before.
>
> We just need disable it for some devices, e.g macvtap and sr-iov.
I think we use expired IP address because  we record IP address in 
virtnet's db, But Now we have reset in it virtnet code.

xutian:
LGTM,

ypu:
LGTM

Thanks @FengYang and @xutian will merge it.

PR #2093, opened 2015-03-23
utils_test.libvirt: Add a function to update polkit rule

will-Do:

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -34,6 +34,7 @@ from virttest import libvirt_storage
 from virttest import utils_net
 from virttest import gluster
 from virttest import remote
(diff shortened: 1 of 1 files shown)


waynesun09:
@will-Do could you use function ```file_replace_append``` in ```LibvirtPolkitConfig``` ?

will-Do:
Yes, I saw that function, and since it can only update content in one line and a new file(rename) used during update, so I rewrite the update step which just use ```re.sub```

will-Do:
Maybe I can rewrite that function then reuse it

waynesun09:
@will-Do no need to update that function, it only used by the _setup_libvirtd function, which could be replaced by utils_config.LibvirtdConfig().
So, ACK for this.

PR #2095, opened 2015-03-24
virtest.tests.unattended_install: Destroy data in iscsi device before in...

FengYang:
...stalling

We need broken the system in iscsi device before installing new windows
system to it.  Or it will not force install the new windows system.
And existed windows system will be used.

id: 1135882


spcui:
LGTM
Acked-by: @spcui

CongSmile:

PR #2097, opened 2015-03-24
Ubuntu 14.04.01 Server i386 (libvirt backend)

igord-daynix:
Fix vm_arch_name configuration parameter for compatibility with kernel arch format

### DIFF
--- a/shared/cfg/guest-os/Linux/Ubuntu/14.04.1-server.i386.cfg
+++ b/shared/cfg/guest-os/Linux/Ubuntu/14.04.1-server.i386.cfg
@@ -1,6 +1,6 @@
 - 14.04.1-server.i386:
     image_name += -14.04.1-server-32
-    vm_arch_name = i386
+    vm_arch_name = i686
     os_variant = ubuntutrusty
     unattended_install, svirt_install:
         kernel = images/ubuntu-server-14-04-1-32/vmlinuz


PandaWei:
Thanks for your work. 
I'm curious about that the kernel arch of ubuntu-14.04.1-server-i386.iso is i686？
Frankly,
The suffix of configuration file's name is i386, and ubuntu-14.04.1-server-i386.iso will be adopted in the installation process.
so i think it's reasonable that vm_arch_name is set to i386.

And, browse the directory shared/cfg/guest-os/Linux/Ubuntu，
We can see that all of vm_arch_name are set to i386 in the other files named xxx-i386.
so keep consistency is good.

igord-daynix:
Thanks for reviewing!
vm_arch_name parameter is used by libvirt backend, you could find in code that this variable is compared with result of platform.machine() method executed on VM under test.
I have tested it on Ubuntu 14.04.1 32bit and platform.machine() method returned 'i686'.

Please validate it for other distributives too.

PandaWei:
Thanks for your reply.
I have checked it. 
Indeed, platform.machine() return 'i686' on RHEL.i386 platform.
but I used 'uname -i', which return 'i386'.
so i think you catch a good and funny problem.


In virttest, we get arch.ARCH by platform.machine().
did you encounter any test fails, which caused by arch.ARCH?

at last,
There are so many vm_arch_name(s) are set to i386 in the other configuration files named xxx-i386.
I don't think it's a good way that just change one of them.
I'm not sure....
maybe we can modify the way to get the architecture of VM. or regard i686 is same as i386.
In fact, both i386 and i686 are all 32-bit Platform.

Other people in ML maybe have better advice...

igord-daynix:
Thanks for your comment.
I have added Ubuntu 14.04.1 guest to virt-test in 1-Feb-2015 and used virt-test/shared/cfg/guest-os/Linux/Ubuntu/14.04-server.i386.cfg for reference.

I caught exception in libvirt_vm.py while trying to run run Ubuntu 14.04.1 VM by libvirt backend:
       support_machine_type = capabs.os_arch_machine_map[hvm_or_pv][arch_name]

I didn't make deep investigations, I found that some other i386 guests use initialize vm_arch_name  by i686. I tried it and it worked.

I found here good explanation regarding i386,i586,i686:
http://www.redhat.com/archives/rhl-list/2005-February/msg05183.html

I think that 32-bit kernels of all virt-test guests is compiled as i686.

ypu:
LGTM

igord-daynix:
Dear Maintainers,
Is any additional action required from me in order to commit this patch.?

Best Regards,

waynesun09:
ACK and merging now

ypu:
@igord-daynix hi normally we will merge a pullreq after it got two acks. And if the second person acked is not the maintainer. We may miss some pullreq

igord-daynix:
Thank You!

PR #2105, opened 2015-03-25
virttest.utils_test: Add option m to the tar command

ypu:
Sometime the different setup in guest and host will cause the timestamp
in the future when we extract auotest package inside guest when we
try to run autotest inside guest. As this is not the test point and we
don't want this to influence our testing, add the option m to the tar
command to ignore this.


ypu:
ID: 1129531

luckyh:
LGTM, acked.

PandaWei:
ACK， 
from my point of view,  good PR.

ypu:
Have a check

ypu:
Thanks @luckyh and @PandaWei will megre it as it passed the Travis test.

PR #2108, opened 2015-03-25
libvirt: Get MAC address of remote VM after it was started

igord-daynix:
In case of ESX protocol, virsh couldn't get MAC address of network adapter before VM was started.


will-Do:
LGTM, ack this patch

yangdongsheng:
Makes sense.

PR #2110, opened 2015-03-26
service.py: Added new method statuslog() to show the service status text

svirt:
New method statuslog() is added. It is needed for getting the status log
text of the service. The existing status() method returns only the status
if the service runs or not

### DIFF
--- a/virttest/staging/service.py
+++ b/virttest/staging/service.py
@@ -129,6 +129,15 @@ def sys_v_init_result_parser(command):
             # If output does not contain a dead flag, check it with "running".
             return bool(re.search(r"running", output))
         return method
+    elif command == "raw_status":
+        def method(cmdResult):
+            """
+            Parse method for service XXX status raw output.
+
+            Returns command result object.
+            """
+            return cmdResult
+        return method
     elif command == "list":
         def method(cmdResult):
             """
@@ -218,6 +227,16 @@ def systemd_result_parser(command):
             # Check it with Active status.
             return (output.count("Active: active") > 0)
         return method
+    elif command == "raw_status":
+        def method(cmdResult):
+            """
+            Parse method for systemctl status XXX raw output.
+
+            Returns command result object.
+            """
+            output = cmdResult
+            return output
+        return method
     elif command == "list":
         def method(cmdResult):
             """
@@ -270,6 +289,9 @@ def sys_v_init_command_generator(command):
     elif command == 'disable':
         command_name = "chkconfig"
         command = "off"
+    elif command == "raw_status":
+        command_name = "chkconfig"
+        command = "status"
     elif command == 'list':
         # noinspection PyUnusedLocal
         def list_command(service_name):
@@ -324,6 +346,8 @@ def systemd_command_generator(command):
         return set_target_command
     elif command == "reset_failed":
         command = "reset-failed"
+    elif command == "raw_status":
+        command = "status"
 
     def method(service_name):
         return [command_name, command, "%s.service" % service_name]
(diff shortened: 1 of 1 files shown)


yangdongsheng:
Is there a use case for this function? It would be great mentioning it here to a reviewer. Thanx

svirt:
Yes, it is used in tests in virsh_managedsave.py, but I still didn't commit the modified tests yet because they depend on this pull request result

virttest/staging/service.py
```diff
@@ -129,6 +129,15 @@ def method(cmdResult):
...
             return bool(re.search(r"running", output))
         return method
+    elif command == "statuslog":
+        def method(cmdResult):
+            """
+            Parse method for service XXX status.
+
+            Returns status log
+            """
+            return cmdResult
```
yangdongsheng: But from the commit message and the name of "statuslog", I am expecting to get a string when I call service.statuslog(), while it's a CmdResult object currently. I am not sure which one is better in your usecase. But I have two suggestion here.

1, change the statuslog to rawstatus or something else, and reword the comments for it. Keep returning a CmdResult object.
2. return cmdresult.output in parser. Keep the comment and commit message.

Either way is acceptable to me, I think. What do you think about it?

Yang

svirt:
After merging this pull request I will be able to commit the use case test

yangdongsheng:
@svirt
Firstly, please address my comment inline, I have shown my concern and my suggestions in that comment.

Secondly, please feel free to send a PR which is depending on this patch. It's better to understand your usecase and to review it.

Thanx.

svirt:
**@yangdongsheng**
Hello, 
yes, I think your mentioned first case is more suitable -  and the name rawstatus looks better, I can rename and rewrite the comments for it.

If it's fine for you, I will do the necessary changes.

Thanks

yangdongsheng:
@svirt
Great!

Please rename it to `raw_status`. others look good to me in this PR.

Thanx

svirt:
**@yangdongsheng**
Yes, I will create a new PR with these changes. Thanks

yangdongsheng:
No need at all. Just update the branch in your local repo and push it to github. Then this PULL REQUEST will be updated automatically.

svirt:
Ok. I will update this branch

svirt:
Looks like there is a problem with running Travis CI on github's side, the log shows:

**fatal: unable to connect to github.com:**
**github.com[0: 192.30.252.129]: errno=Connection timed out**

more details are in the log

svirt:
closed/reopened the PR  to run Travis CI again...

waynesun09:
@svirt could you squash your two commits into one, thx

svirt:
**@waynesun09**
yes sure,done. Now there is one commit only

waynesun09:
ACK
@svirt @yangdongsheng Thanks! merging now.

virttest/staging/service.py
```diff
@@ -270,6 +289,9 @@ def sys_v_init_command_generator(command):
     elif command == 'disable':
         command_name = "chkconfig"
         command = "off"
+    elif command == "raw_status":
+        command_name = "chkconfig"
+        command = "status"
```
yangdongsheng: Why the command_name is chkconfig?

yangdongsheng: I guess you did not test this patch on a sysV environment yet. I think the command_name should be service in here. Do you mind send another patch for it?

svirt:
**@yangdongsheng**
Hi, yes I agree, the command_name = "service" is defined above already and command_name = "chkconfig" is not needed .
I will remove that line. Shell I request a new pull request?

yangdongsheng:
Yes, as there is no need to rename the command_name and the command, I think this branch is whole not needed. It's better to do a test for it I suppose on a sys_v style environment. Thanx for your work.

svirt:
**@yangdongsheng**
Thanks for the review, I will do some checks and then prepare another pull request

PR #2117, opened 2015-03-30
shared.unattended: Updating RHEL-6-spice.ks for RHEL 6.7

swapnakrishnan2k:
Autotest generates a random mac address for the nic device. Udev rules
will generate a new interface for each "new" device when guest boots
up. So the nic interface changes and this confuses the network manager
inside guest which leads to system will not get "new" interface get IP
address automatically.

To avoid this, we comment out the rule being created with each boot.
This will allow the device to get IP address automatically.

Reviewed By: Swapna Krishnan <skrishna@redhat.com>


ypu:
LGTM

PR #2121, opened 2015-04-01
libvirt_vm: Compatible qemu-guest-agent service for RHEL6&7

will-Do:

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -2383,16 +2383,35 @@ class VM(virt_vm.BaseVM):
 
         def _is_ga_finished():
             return (session.cmd_status("pgrep qemu-ga") == 1)
+
+        def _start_ga():
+            if not _is_ga_running():
+                cmd = "service qemu-guest-agent start"
+                status, output = session.cmd_status_output(cmd)
+                if status and "unrecognized service" in output:
+                    cmd = "service qemu-ga start"
+                    status, output = session.cmd_status_output(cmd)
+                if status:
+                    raise virt_vm.VMError("Start qemu-guest-agent failed:"
+                                          "\n%s" % output)
+
+        def _stop_ga():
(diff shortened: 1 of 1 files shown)


Hao-Liu:
Good. ACK this.

virttest/libvirt_vm.py
```diff
@@ -2383,16 +2383,35 @@ def _is_ga_running():
...
+        def _start_ga():
+            if not _is_ga_running():
+                cmd = "service qemu-guest-agent start"
+                status, output = session.cmd_status_output(cmd)
+                if status and "unrecognized service" in output:
+                    cmd = "service qemu-ga start"
+                    status, output = session.cmd_status_output(cmd)
+                if status:
+                    raise virt_vm.VMError("Start qemu-guest-agent failed:"
+                                          "\n%s" % output)
```
PandaWei: Good PR,
but there's something wrong with the last if-block.
"service qemu-ga start" mabybe fail,
the error message hints that "Start qemu-guest-agent failed".

Just a suggestion.
It's better that we should determine which one is supported at first.
and then call session.cmd_status_output() to start or stop it.

will-Do: Not really. 
The error message prompt here is not corresponding to the service name 100%(Actually, the package name is qemu-guest-agent). If 'service qemu-ga start' fail, both 'status' and 'output' will be overwirted, and that's what I want.
Determine 'qemu-ga' or 'qemu-guest-agent' is unnecessary and not convenient for different platform. Just run the service command and check the output is enough.

PandaWei: OK,
Thanks your explanation,
I have no other advice.
leave ACK

chuanchang:
LGTM, ACKed.

will-Do:
Thx @Hao-Liu @PandaWei @chuanchang, merge this PR

PR #2122, opened 2015-04-01
utils_test/libvirt: add check point for iface_check

PandaWei:
Add check point to check the state of specific interface.
return True if state is active, otherwise return False.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1255,8 +1255,10 @@ def check_iface(iface_name, checkpoint, extra="", **dargs):
     Check interface with specified checkpoint.
 
     :param iface_name: Interface name
-    :param checkpoint: Check if interface exists, MAC address, IP address or
-                       ping out. Support values: [exists, mac, ip, ping]
+    :param checkpoint: Check if interface exists,
+                       and It's MAC address, IP address and State,
+                       also connectivity by ping.
+                       valid checkpoint: [exists, mac, ip, ping, state]
     :param extra: Extra string for checking
     :return: Boolean value, true for pass, false for fail
     """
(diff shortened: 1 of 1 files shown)


PandaWei:
refer to autotest/tp-libvirt#510

virttest/utils_test/libvirt.py
```diff
@@ -1291,6 +1293,18 @@ def check_iface(iface_name, checkpoint, extra="", **dargs):
             iface_ip = iface.get_ip()
             check_pass = iface_ip == extra
             logging.debug("IP address of %s: %s", iface_name, iface_ip)
+        elif checkpoint == "state":
+            # check iface State
+            result = virsh.iface_list(extra, ignore_status=True)
+            check_exit_status(result, False)
+            output = re.findall(r"(\S+)\ +(\S+)\ +(\S+|\s+)[\ +\n]",
+                                str(result.stdout))
+            iface_state = filter(lambda x: x[0] == iface_name, output[1:])[0][1]
```
will-Do: It's better to deal with re and filter in a 'try...except' block

virttest/utils_test/libvirt.py
```diff
@@ -1291,6 +1293,18 @@ def check_iface(iface_name, checkpoint, extra="", **dargs):
...
+            # active corresponds True, otherwise return False
+            if iface_state is not None and iface_state == "active":
+                check_pass = True
+            else:
+                check_pass = False
```
will-Do: ```check_pass = iface_state == "active"``` should be enough

PandaWei: Hi, 
Thanks for your comments.
I think it's very useful suggestion.
I will update this PR as it.

virttest/utils_test/libvirt.py
PandaWei: check_iface() has a big try ... except that wraps up all of if and elif block.

will-Do: Thanks, then I'm good for this PR

yangdongsheng:
Fine to me.

Hao-Liu:
LGTM,  Merging. Thanks @PandaWei @will-Do @yangdongsheng .

PR #2125, opened 2015-04-03
Two update for get_deps_dir in date_dir

ypu:
Current get_deps_dir() can not work under python 2.4 env as python have some bug in that version. So make a work around for it. And also update the function for checking deps dirs as deps is under shared in virt-test, the original check will never find it.

### DIFF
--- a/virttest/data_dir.py
+++ b/virttest/data_dir.py
@@ -182,10 +182,13 @@ def get_deps_dir(target=None):
     # Get the frame that called this function
     frame = inspect.stack()[1]
     # This is the module that called the function
-    module = inspect.getmodule(frame[0])
     # With the module path, we can keep searching with a parent dir with 'deps'
     # in it, which should be the correct deps directory.
-    path = os.path.dirname(module.__file__)
+    try:
+        module = inspect.getmodule(frame[0])
+        path = os.path.dirname(module.__file__)
+    except TypeError:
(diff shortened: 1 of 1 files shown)


ypu:
ID: 1207544

virttest/data_dir.py
```diff
@@ -182,13 +182,21 @@ def get_deps_dir(target=None):
...
     frame = inspect.stack()[1]
     # This is the module that called the function
-    module = inspect.getmodule(frame[0])
+    try:
+        module = inspect.getmodule(frame[0])
+        path = os.path.dirname(module.__file__)
+    except TypeError:
+        path = os.path.dirname(frame[1])
     # With the module path, we can keep searching with a parent dir with 'deps'
     # in it, which should be the correct deps directory.
```
PandaWei: I think it's better to move the comments before module = inspect.getmodule(frame[0]) .
Others looks Good.
ACK.

XueqiangWei:
LGTM,
Acked by Xueqiang Wei <xuwei@redhat.com>

xiagao:
Acked-by: Xiaoling Gao <xiagao@redhat.com>

ypu:
Thanks @PandaWei move the comments to above

ypu:
Thanks @PandaWei @xiaoqing-wei @xiagao for your reviewing. As the second update only update some comments. So will mere it.

PR #2127, opened 2015-04-07
util_test/libvirt: raise TestNAError instead of TestError

PandaWei:
TestNAError seems more reasonable.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -505,7 +505,7 @@ def setup_or_cleanup_gluster(is_setup, vol_name, brick_path="", pool_name="",
     try:
         utils_misc.find_command("gluster")
     except ValueError:
-        raise error.TestError("Missing command 'gluster'")
+        raise error.TestNAError("Missing command 'gluster'")
     if not brick_path:
         tmpdir = os.path.join(data_dir.get_root_dir(), 'tmp')
         brick_path = os.path.join(tmpdir, pool_name)


virttest/utils_test/libvirt.py
```diff
@@ -505,7 +505,7 @@ def setup_or_cleanup_gluster(is_setup, vol_name, brick_path="", pool_name="",
     try:
         utils_misc.find_command("gluster")
     except ValueError:
-        raise error.TestError("Missing command 'gluster'")
+        raise error.TestNAError("Missing command 'gluster'")
```
will-Do: We try to decrease the SKIP cases, and 'gluster' is a big part of our test cases, so we hope the case should fail if 'gluster' command not installed and let the tester install and run the test again.

PandaWei: Thanks for your response.
yeah, 
Maybe your intent is good.
But it doesn't sounds reasonable and convincing.
Missing command should be classified as Test Environment Problem which Do not meet the requirements.
It's not failure.
and all of this kind of excepts are regarded as TestNAError in Virttest.
we should use uniform principle, 
it's need to take more care for it, because it's common API module.
and I think it's a bad idea to treat 'Missing gluster' as a especial case than others.
 
consider your situation,  Maybe you can output reminder information at cases for user.

will-Do:
Fair enough, then no problem from me. ACK

Hao-Liu:
Agree on keep using TestNAError when test environment doesn't meet requirement. Merge this. Thanks @PandaWei @will-Do.

PR #2128, opened 2015-04-08
virttest: Turn off export of volume through NFS

chuanchang:
NFS should not be used to support the lock volume used by CTDB as NFS does not currently support locking. Instead, use the Gluster native mount. 

### DIFF
--- a/virttest/gluster.py
+++ b/virttest/gluster.py
@@ -335,3 +335,24 @@ def add_rpc_insecure(filepath):
         cmd = "sed -i '/end-volume/i \ \ \ \ option rpc-auth-allow-insecure on' %s" % filepath
         utils.system(cmd)
         utils.system("service glusterd restart; sleep 2")
+
+
+@error.context_aware
+def gluster_nfs_disable(vol_name):
+    """
+    Turn-off export of volume through NFS
+    :param vol_name: name of gluster volume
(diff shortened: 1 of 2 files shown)


virttest/gluster.py
```diff
@@ -335,3 +335,24 @@ def add_rpc_insecure(filepath):
         cmd = "sed -i '/end-volume/i \ \ \ \ option rpc-auth-allow-insecure on' %s" % filepath
         utils.system(cmd)
         utils.system("service glusterd restart; sleep 2")
+
+
+@error.context_aware
+def gluster_nfs_disable(vol_name):
+    """
+    Turn-off export of volume through NFS
```
will-Do: Usually, a blank line needed here

virttest/gluster.py
```diff
@@ -335,3 +335,24 @@ def add_rpc_insecure(filepath):
...
+    utils.system(cmd)
+
+    cmd = "gluster volume info"
+    output = utils.system_output(cmd)
+    match = re.findall(r'nfs.disable: on', output)
+
+    if not match:
+        return False
+    else:
+        return True
```
will-Do: Maybe ```return bool(match)``` is enough

lmr:

Thanks!

PR #2130, opened 2015-04-09
virttest/test_setup: make sure all pids killed

PandaWei:
Sleep 3 sec is not enough sometimes,
It's necessary to ensure all of pids end up before follow-up steps.

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -1675,7 +1675,13 @@ class EGDConfig(object):
             for pid in self.env.data["egd_pids"]:
                 logging.info("Stop egd.pl(%s)" % pid)
                 utils.signal_pid(pid, 15)
-            # give time to wait port released by egd.pl
-            time.sleep(3)
+
+            def _all_killed():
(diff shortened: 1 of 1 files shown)


virttest/test_setup.py
```diff
@@ -1675,7 +1675,16 @@ def cleanup(self):
...
-            # give time to wait port released by egd.pl
-            time.sleep(3)
+
+            def _all_killed():
+                for pid in self.env.data["egd_pids"]:
+                    if utils.pid_is_alive(pid):
+                        return False
+                return True
+            # wait port released by egd.pl
+            if utils.wait_for(_all_killed, timeout=60) is not True:
```
Hao-Liu: Why bother wait another 3 seconds if you've already waited 1 minute? I think just `utils.wait_for(_all_killed, timeout=60)` should be okay.

PandaWei: Thanks for your response and comment.
updated it.

Hao-Liu:
LGTM

will-Do:
Good

PR #2132, opened 2015-04-10
virttest/test_setup: check if hugepages is supported

PandaWei:
It's better to check whether the system does support hugepages
function as same as the process for transparent hugepage.
This commit references the validating method in class libhugetlbfs
that declared at autotest/client/tests/libhugetlbfs/libhugetlbfs.py

### DIFF
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -54,6 +54,14 @@ class THPKhugepagedError(THPError):
     pass
 
 
+class HPNotSupportedError(Exception):
+
+    """
+    Thrown when host does not support hugepages.
+    """
+    pass
+
+
 class PolkitConfigError(Exception):
 
     """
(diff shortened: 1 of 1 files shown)


virttest/test_setup.py
```diff
@@ -286,7 +294,10 @@ def __init__(self, params):
         self.qemu_overhead = int(params.get("hugepages_qemu_overhead", 128))
         self.deallocate = params.get("hugepages_deallocate", "yes") == "yes"
         self.hugepage_path = '/mnt/kvm_hugepage'
-        self.kernel_hp_file = '/proc/sys/vm/nr_hugepages'
+        if os.path.exists('/proc/sys/vm/nr_hugepages'):
+            self.kernel_hp_file = '/proc/sys/vm/nr_hugepages'
+        else:
+            raise HPNotSupportedError("System doesn't support hugepages")
```
yangdongsheng: Is this only required by hugepages? what about the others in this section?

PandaWei: Thanks for your reminder.
The transparent hugepage and hugepage class were declared at this module.
but  THPNotSupportedError  has been considered.
so we can don't worry about it.

will-Do:
@PandaWei 
You need rebase this patch before we merge it

PandaWei:
Thanks,
update it

waynesun09:
ACK and merging now.

PR #2133, opened 2015-04-10
virttest: the params of docstring should begin with colon

PandaWei:
This patch does what is the same as commit de30fa4.


Hao-Liu:
Ack and merging. Thanks @PandaWei

PR #2135, opened 2015-04-10
utils_test.libvirt: Support prepare persistent network

will-Do:

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -95,10 +95,15 @@ class LibvirtNetwork(object):
         ip = utils_net.get_ip_address_by_interface(self.iface)
         return ip, net_xml
 
-    def __init__(self, net_type, address=None, iface=None):
-        self.name = 'virt-test-%s' % net_type
+    def __init__(self, net_type, address=None, iface=None, net_name=None,
+                 persistent=False):
+        if net_name is None:
+            self.name = 'virt-test-%s' % net_type
+        else:
+            self.name = net_name
         self.address = address
         self.iface = iface
+        self.persistent = persistent
 
         if net_type == 'vnet':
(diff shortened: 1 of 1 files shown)


yangdongsheng:
@Hao-Liu @will-Do You are sending two similar PR for this functionality.

Hao-Liu:
I've close that PR and will move to this one.

virttest/utils_test/libvirt.py
```diff
@@ -114,12 +119,26 @@ def __init__(self, net_type, address=None, iface=None):
...
+                        "Please use cleanup() instead.")
+        return virsh.net_destroy(self.name)
+
+    def cleanup(self):
+        """
+        Clear up network.
+        """
+        if self.persistent:
+            virsh.net_undefine(self.name)
         return virsh.net_destroy(self.name)
```
Hao-Liu: We should destroy the net before undefine. It will fail if we undefining a live network on RHEL6.

virttest/utils_test/libvirt.py
```diff
@@ -114,12 +119,26 @@ def __init__(self, net_type, address=None, iface=None):
...
+            net_xml.define()
+            net_xml.start()
+        else:
+            net_xml.create()
 
     def destroy(self):
         """
         Clear up created network.
         """
+        logging.warning("This function was deprecated. "
```
Hao-Liu: You can remove this function, since this class is now only used by graphics test and I'm debugging it. I'll update to not use this function.

will-Do:
@Hao-Liu 
Thanks, just update it

waynesun09:
ACK

waynesun09:
@will-Do @yangdongsheng @Hao-Liu Thanks for the update, merging now.

PR #2139, opened 2015-04-13
virttest: Add support for ceph

ypu:
Make guest can use ceph backends.

### DIFF
--- a/shared/cfg/guest-hw.cfg
+++ b/shared/cfg/guest-hw.cfg
@@ -220,6 +220,7 @@ variants:
 variants image_backend:
     - @filesystem:
     - gluster:
+        storage_type = glusterfs
         enable_gluster = yes
         # please go through https://github.com/autotest/virt-test/wiki/Virt-test-support-for-GlusterFs for more details.
         #Set the brick of gluster server. If set, will try to set gluster
@@ -249,6 +250,7 @@ variants image_backend:
         #lvm_reload_cmd += "systemctl reload lvm2-monitor.service"
         force_remove_image = no
     - emulated_lvm:
+        storage_type = emulated lvm
         only raw
         force_create_image = no
         image_raw_device = yes
@@ -261,6 +263,16 @@ variants image_backend:
         remove_emulated_image = no
         #lvm_reload_cmd = "systemctl reload lvm2-lvmetad.service;"
         #lvm_reload_cmd += "systemctl reload lvm2-monitor.service"
+    - ceph:
+        storage_type = ceph
+        # Some test case may need images from different backend. Please set up
+        # enable_ceph_$image_name to no if the special image is not use ceph
+        # server in your test
+        enable_ceph = yes
+        cmds_installed_host += " rbd"
+        # Please update following parameters based on your test env
+        #ceph_monitor = $ceph_monitor_ip
+        #rbd_pool_name = $autotest_pool_name
 
 variants:
     - @no_virtio_rng:
--- a/virttest/ceph.py
+++ b/virttest/ceph.py
@@ -1,0 +1,157 @@
+"""
+CEPH Support
+This file has the functions that helps
+* To create rbd pool
+* To map/unmap rbd pool
+* To mount/umount cephfs to localhost
+* To return rbd uri which can be used as disk image file path.
+"""
+
+import logging
+import os
+import re
+from autotest.client.shared import utils
+from autotest.client.shared import error
+import utils_misc
+
+
+class CephError(Exception):
+    pass
+
+
+@error.context_aware
+def rbd_image_create(ceph_monitor, rbd_pool_name, rbd_image_name,
+                     rbd_image_size, force_create=False):
+    """
+    Create a rbd image.
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params rbd_image_name: The name of rbd image
+    :params rbd_image_size: The size of rbd image
+    :params force_create: Force create the image or not
+    """
+    if rbd_image_exist(ceph_monitor, rbd_pool_name, rbd_image_name):
+        create_image = False
+        image_info = rbd_image_info(ceph_monitor, rbd_pool_name,
+                                    rbd_image_name)
+        try:
+            int(rbd_image_size)
+            compare_str = rbd_image_size
+        except ValueError:
+            compare_str = utils_misc.normalize_data_size(rbd_image_size)
+        if image_info['size'] != compare_str or force_create:
+            rbd_image_rm(ceph_monitor, rbd_pool_name, rbd_image_name)
+            create_image = True
+    if create_image:
+        cmd = "rbd create %s/%s -m %s" % (rbd_pool_name, rbd_image_name,
+                                          ceph_monitor)
+        utils.system(cmd, verbose=True)
+    else:
+        logging.debug("Image already exist skip the create.")
+
+
+@error.context_aware
+def rbd_image_rm(ceph_monitor, rbd_pool_name, rbd_image_name):
+    """
+    Remove a rbd image
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params rbd_image_name: The name of rbd image
+    """
+    if rbd_image_exist(ceph_monitor, rbd_pool_name, rbd_image_name):
+        cmd = "rbd rm %s/%s -m %s" % (rbd_pool_name, rbd_image_name,
+                                      ceph_monitor)
+        utils.system(cmd, verbose=True)
+    else:
+        logging.debug("Image not exist, skip to remove it.")
+
+
+@error.context_aware
+def rbd_image_exist(ceph_monitor, rbd_pool_name, rbd_image_name):
+    """
+    Check if rbd image is exist
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params rbd_image_name: The name of rbd image
+    """
+    cmd = "rbd ls %s -m %s" % (rbd_pool_name, ceph_monitor)
+    output = utils.system_output(cmd, ignore_status=True, verbose=True)
+
+    logging.debug("Resopense from rbd ls command is: %s" % output)
+
+    return (rbd_image_name.strip() in output.splitlines())
+
+
+@error.context_aware
+def rbd_image_info(ceph_monitor, rbd_pool_name, rbd_image_name):
+    """
+    Get information of a rbd image
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params rbd_image_name: The name of rbd image
+    """
+    cmd = "rbd info %s/%s -m %s" % (rbd_pool_name, rbd_image_name,
+                                    ceph_monitor)
+    output = utils.system(cmd)
+    info_pattern = "rbd image \'%s\':.*?$" % rbd_image_name
+
+    rbd_image_info_str = re.findall(info_pattern, output, re.S)[0]
+
+    rbd_image_info = {}
+    for rbd_image_line in rbd_image_info_str.splitlines():
+        if ":" not in rbd_image_line:
+            if "size" in rbd_image_line:
+                size_str = re.findall("size\s+(\d+\s+\w+)\s+",
+                                      rbd_image_line)[0]
+                size = utils_misc.normalize_data_size(size_str)
+                rbd_image_info['size'] = size
+            if "order" in rbd_image_line:
+                rbd_image_info['order'] = int(re.findall("order\s+(\d+)",
+                                                         rbd_image_line))
+        else:
+            tmp_str = rbd_image_line.strip().split(":")
+            rbd_image_info[tmp_str[0]] = tmp_str[1]
+    return rbd_image_info
+
+
+@error.context_aware
+def rbd_image_map(ceph_monitor, rbd_pool_name, rbd_image_name):
+    """
+    Maps the specified image to a block device via rbd kernel module
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params rbd_image_name: The name of rbd image
+    """
+    cmd = "rbd map %s --pool %s -m %s" % (rbd_image_name, rbd_pool_name,
+                                          ceph_monitor)
+    output = utils.system_output(cmd, verbose=True)
+    if os.path.exist(os.path.join("/dev/rbd", rbd_pool_name, rbd_image_name)):
+        return os.path.join("/dev/rbd", rbd_pool_name, rbd_image_name)
+    else:
+        logging.debug("Failed to map image to local: %s" % output)
+        return None
+
+
+@error.context_aware
+def rbd_image_unmap(rbd_pool_name, rbd_image_name):
+    """
+    Unmaps the block device that was mapped via the rbd kernel module
+    :params rbd_pool_name: The name of rbd pool
+    :params rbd_image_name: The name of rbd image
+    """
+    cmd = "rbd unmap /dev/rbd/%s/%s" % (rbd_pool_name, rbd_image_name)
+    output = utils.system_output(cmd, verbose=True)
+    if os.path.exist(os.path.join("/dev/rbd", rbd_pool_name, rbd_image_name)):
+        logging.debug("Failed to unmap image from local: %s" % output)
+
+
+@error.context_aware
+def get_image_filename(ceph_monitor, rbd_pool_name, rbd_image_name):
+    """
+    Return the rbd image file name
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params rbd_image_name: The name of rbd image
+    """
+    return "rbd:%s/%s:mon_host=%s" % (rbd_pool_name, rbd_image_name,
+                                      ceph_monitor)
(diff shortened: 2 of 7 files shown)


ypu:
ID: 1204072

virttest/ceph.py
```diff
@@ -0,0 +1,155 @@
...
+        logging.debug("Failed to unmap image from local: %s" % output)
+
+
+@error.context_aware
+def get_image_filename(ceph_monitor, rbd_pool_name, rbd_image_name):
+    """
+    Return the rbd image file name
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params ceph_image_name: The name of rbd image
```
spcui: params ceph_image_name should be "rbd_image_name".

virttest/ceph.py
```diff
@@ -0,0 +1,155 @@
...
+        return None
+
+
+@error.context_aware
+def rbd_image_unmap(rbd_pool_name, rbd_image_name):
+    """
+    Unmaps the block device that was mapped via the rbd kernel module
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params ceph_image_name: The name of rbd image
```
spcui: params ceph_image_name should be "rbd_image_name".

virttest/ceph.py
```diff
@@ -0,0 +1,155 @@
...
+    else:
+        logging.debug("Failed to map image to local: %s" % output)
+        return None
+
+
+@error.context_aware
+def rbd_image_unmap(rbd_pool_name, rbd_image_name):
+    """
+    Unmaps the block device that was mapped via the rbd kernel module
+    :params ceph_monitor: The specified monitor to connect to
```
spcui: need delete this line.

virttest/ceph.py
```diff
@@ -0,0 +1,155 @@
...
+    return rbd_image_info
+
+
+@error.context_aware
+def rbd_image_map(ceph_monitor, rbd_pool_name, rbd_image_name):
+    """
+    Maps the specified image to a block device via rbd kernel module
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params ceph_image_name: The name of rbd image
```
spcui: params ceph_image_name should be "rbd_image_name".

virttest/ceph.py
```diff
@@ -0,0 +1,155 @@
...
+    return (rbd_image_name.strip() in output.splitlines())
+
+
+@error.context_aware
+def rbd_image_info(ceph_monitor, rbd_pool_name, rbd_image_name):
+    """
+    Get information of a rbd image
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params ceph_image_name: The name of rbd image
```
spcui: params ceph_image_name should be "rbd_image_name".

virttest/ceph.py
```diff
@@ -0,0 +1,155 @@
...
+        logging.debug("Image already exist skip the create.")
+
+
+@error.context_aware
+def rbd_image_rm(ceph_monitor, rbd_pool_name, rbd_image_name):
+    """
+    Remove a rbd image
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params ceph_image_name: The name of rbd image
```
spcui: params ceph_image_name should be "rbd_image_name".

virttest/ceph.py
```diff
@@ -0,0 +1,155 @@
...
+        logging.debug("Image not exist, skip to remove it.")
+
+
+@error.context_aware
+def rbd_image_exist(ceph_monitor, rbd_pool_name, rbd_image_name):
+    """
+    Check if rbd image is exist
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params ceph_image_name: The name of rbd image
```
spcui: params ceph_image_name should be "rbd_image_name".

virttest/ceph.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+@error.context_aware
+def rbd_image_create(ceph_monitor, rbd_pool_name, rbd_image_name,
+                     rbd_image_size, force_create=False):
+    """
+    Create a rbd image.
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params ceph_image_name: The name of rbd image
```
spcui: params ceph_image_name should be "rbd_image_name".

ypu:
Thanks @spcui will update the docstrings.

shared/cfg/guest-hw.cfg
```diff
@@ -259,6 +259,11 @@ variants image_backend:
         remove_emulated_image = no
         #lvm_reload_cmd = "systemctl reload lvm2-lvmetad.service;"
         #lvm_reload_cmd += "systemctl reload lvm2-monitor.service"
+    - ceph:
+        enable_ceph = yes
+        # Please update following parameters based on your test env
+        # ceph_monitor = $ceph_monitor_ip
+        # rbd_pool_name = $autotest_pool_name
```
suqinhuang: Can we drop the blank space before ceph_monitor and rbd_pool_name, the params invalid at first, as I only remove "#", it may happen to others as well.

virttest/ceph.py
```diff
@@ -0,0 +1,155 @@
...
+        image_info = rbd_image_info(ceph_monitor, rbd_pool_name,
+                                    rbd_image_name)
+        try:
+            int(rbd_image_size)
+            compare_str = rbd_image_size
+        except ValueError:
+            compare_str = utils_misc.normalize_data_size(rbd_image_size)
+        if image_info['size'] != compare_str or force_create:
+            rbd_image_rm(ceph_monitor, rbd_pool_name, rbd_image_name)
+            create_image = True
```
suqinhuang: script doesn't work when set force_create_image = yes

suqinhuang: installation doesn't work because rbd is used to iso as well

    -drive id=drive_cd1,if=none,snapshot=off,aio=native,media=cdrom,file=rbd:qemu-kvm-pool/RHEL6.7-Server-x86_64.iso.raw:mon_host=10.66.106.22 \
    -device ide-cd,id=cd1,drive=drive_cd1,bootindex=2,bus=ide.0,unit=0 \
    -drive id=drive_unattended,if=none,snapshot=off,aio=native,media=cdrom,file=rbd:qemu-kvm-pool/ks.iso.raw:mon_host=10.66.106.22 \
    -device ide-cd,id=unattended,drive=drive_unattended,bootindex=3,bus=ide.0,unit=1 \

shared/cfg/guest-hw.cfg
suqinhuang: some idea:
How about set ramdom name of the pool? then ceph can be tested by severl people at the same time. otherwise the guest is overwrite if they test the same guest.

virttest/ceph.py
suqinhuang: others:
Didn't find following error prompt in .debug file when librbd is not installed, which can be shown manually

qemu-img: error loading librbd: /usr/lib64/qemu/librbd.so.1: cannot open shared object file: No such file or directory
qemu-img: rbd:qemu-kvm-pool/rhel71-64-virtio.raw:mon_host=10.66.106.22: Could not create image: Input/output error

shared/cfg/guest-hw.cfg
ypu: Hi @suqinhuang 
Above options must setup based on the test env. So there is no reason to set up some fake value for them. And the docstring already told the user to file them based on the test env. So I think it is OK. 

And for the random name idea, as we talked before it not only needed by ceph in our env. NFS and iscsi may also need this solution. So we'd better do it in separated pullreq which may the task clean and easier to cover and design it.

virttest/ceph.py
ypu: @suqinhuang hi 
1. I tested it with force_create with unatteded_isntall seems it works.
2. For unattedned_install you need make enable_ceph for cdroms  to no just like gluster fs. 
3. Checked my host even after install librbd1 there is still don't have librbd.so.1 under /usr/lib64/qemu. Can you borrow your test env for check it? It is under /usr/lib64 in my host. Can you help me to check this is a bug or not?

shared/cfg/guest-hw.cfg
suqinhuang: Hi ypu,
point it out because no blank in other params, following are the exapmles

1. 
    # following lines to enable the vhost support ( only available
        # for tap )
        #netdev_extra_params = ",vhost=on"

2. 
        storage_type = lvm
        emulational_device = yes
        # specify a file as emulated volume, if not will create file in
        # autotest data dir;
        #emulated_image =
        force_remove_image = no
        remove_emulated_image = no
        #lvm_reload_cmd = "systemctl reload lvm2-lvmetad.service;"
        #lvm_reload_cmd += "systemctl reload lvm2-monitor.service"

virttest/ceph.py
suqinhuang: librbd is installed in my host, but didn't find the error in the debug file.
What I confused here is: why no error prompt for the package miss.

For force_create i will double check.

suqinhuang: seems the image is not removed:

/bin/qemu-img create -f raw rbd:qemu-kvm-pool/rhel70-64-virtio.raw:mon_host=10.66.106.22 20G
Formatting 'rbd:qemu-kvm-pool/rhel70-64-virtio.raw:mon_host=10.66.106.22', fmt=raw size=21474836480
qemu-img: rbd:qemu-kvm-pool/rhel70-64-virtio.raw:mon_host=10.66.106.22: Could not create image: File exists

shared/cfg/guest-hw.cfg
ypu: OK thanks for your explain. I think I miss understand your request. And OK I will update this part.

virttest/ceph.py
ypu: Thanks for @suqinhuang to share the test env. And after check seems this caused by missing command of rbd. So will add command check for rbd.

lmr:
@ypu I've removed an old autotest package from my PPA, only to realize that virt-test was using it. Anyway, I've fixed the problem, my apologies for the inconvenience, please rebase this PR to the latest master. Thanks!

ypu:
@lmr Thanks for helping to resolve this. pullreq updated.

ypu:
@suqinhuang and @spcui can you help to review it again?

virttest/ceph.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    Remove a rbd image
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params rbd_image_name: The name of rbd image
+    """
+    if rbd_image_exist(ceph_monitor, rbd_pool_name, rbd_image_name):
+        cmd = "rbd rm %s/%s -m %s" % (rbd_pool_name, rbd_image_name,
+                                      ceph_monitor)
+        utils.system(cmd)
```
suqinhuang: CmdError: Command <rbd rm qemu-kvm-pool/win7-64-sp1-virtio.raw -m 10.66.106.22> failed, rc=16, Command returned non-zero exit status    [context: preprocessing --> (rbd_image_rm)]
* Command:
    rbd rm qemu-kvm-pool/win7-64-sp1-virtio.raw -m 10.66.106.22
Exit status: 16
Duration: 0.0533900260925


I can remove it manually, need to sleep for a while ?

virttest/ceph.py
suqinhuang: Hi ypu,
I think need to make ceph working for unattend_intall in upstream as well.

Thanks

suqinhuang:
Some other comment
1. I think need to make ceph works for unattended_install in upstream as well
2. ceph is raw only
3. copy will fail when installation fail (let's ignore image copy for ceph?)
4. suggest to set "force_create_image = yes" as default

virttest/ceph.py
ypu: Hi
I think unattended_install should works, what problem you met?

virttest/ceph.py
ypu: perror 16
OS error code  16:  Device or resource busy
Not met this during my test. Will confirm with feature owner about this.

virttest/ceph.py
```diff
@@ -0,0 +1,154 @@
...
+        try:
+            int(rbd_image_size)
+            compare_str = rbd_image_size
+        except ValueError:
+            compare_str = utils_misc.normalize_data_size(rbd_image_size)
+        if image_info['size'] != compare_str or force_create:
+            rbd_image_rm(ceph_monitor, rbd_pool_name, rbd_image_name)
+            create_image = True
+    if create_image:
+        cmd = "rbd create %s/%s -m %s" % (rbd_pool_name, rbd_image_name,
```
suqinhuang: Can we add debug info here for rbd creation prompt

virttest/ceph.py
```diff
@@ -0,0 +1,154 @@
...
+
+@error.context_aware
+def rbd_image_rm(ceph_monitor, rbd_pool_name, rbd_image_name):
+    """
+    Remove a rbd image
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params rbd_image_name: The name of rbd image
+    """
+    if rbd_image_exist(ceph_monitor, rbd_pool_name, rbd_image_name):
```
suqinhuang: can we add debug here for rbd rm prompt

virttest/ceph.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+@error.context_aware
+def rbd_image_exist(ceph_monitor, rbd_pool_name, rbd_image_name):
+    """
+    Check if rbd image is exist
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params rbd_image_name: The name of rbd image
+    """
```
suqinhuang: can we add debug info here for rbd ls check

virttest/ceph.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+@error.context_aware
+def rbd_image_map(ceph_monitor, rbd_pool_name, rbd_image_name):
+    """
+    Maps the specified image to a block device via rbd kernel module
+    :params ceph_monitor: The specified monitor to connect to
+    :params rbd_pool_name: The name of rbd pool
+    :params rbd_image_name: The name of rbd image
+    """
```
suqinhuang: can we add debug info here for rbd map prompt

virttest/ceph.py
```diff
@@ -0,0 +1,154 @@
...
+        return None
+
+
+@error.context_aware
+def rbd_image_unmap(rbd_pool_name, rbd_image_name):
+    """
+    Unmaps the block device that was mapped via the rbd kernel module
+    :params rbd_pool_name: The name of rbd pool
+    :params rbd_image_name: The name of rbd image
+    """
```
suqinhuang: can we add debug info here for rbd unmap prompt

ypu:
@suqinhuang already update the log and comments, can you help to review it again.

virttest/ceph.py
```diff
@@ -0,0 +1,155 @@
...
+            compare_str = rbd_image_size
+        except ValueError:
+            compare_str = utils_misc.normalize_data_size(rbd_image_size)
+        if image_info['size'] != compare_str or force_create:
+            rbd_image_rm(ceph_monitor, rbd_pool_name, rbd_image_name)
+            create_image = True
+    if create_image:
+        cmd = "rbd create %s/%s -m %s" % (rbd_pool_name, rbd_image_name,
+                                          ceph_monitor)
+        utils.system(cmd, verbose=True)
```
suqinhuang: In my understand, the cmd will be printed in .debug file when verbose=True

I mean can we add description for env prepare in .info file 
e.g here  logging.info("Create a rbd image")
welcome your feedback, if something I mis-understand

ypu: OK, you just said add some debug log here. So I think you mean add something in debug log file. Not get your meaning at the first place. But if we add all place you marked in the file,  is it too much for info log file?

suqinhuang:
Acked-by: Suqin Huang

spcui:
LGTM
Acked-by: scui@redhat.com

suqinhuang:
Acked-by: Suqin Huang

xutian:
@ypu  we still need to update functions  check, info, remove  for qemu_storage.QemuImg class.

Thanks,
@xutian

shared/cfg/guest-hw.cfg
```diff
@@ -220,6 +220,7 @@ variants:
 variants image_backend:
     - @filesystem:
     - gluster:
+        image_backend = glusterfs
         enable_gluster = yes
```
xutian: it looks we have param use_storage to mark different image backend why not use it.

virttest/env_process.py
```diff
@@ -306,6 +312,7 @@ def postprocess_image(test, params, image_name, vm_process_status=None):
     if params.get("restore_image_after_testing", "no") == "yes":
         image.backup_image(params, base_dir, "restore", True)
     if params.get("remove_image") == "yes":
+        logging.info("Remove image on %s." % image.image_backend)
         if clone_master is None:
             image.remove()
```
xutian: it looks we need call  ceph.rbd_image_rm function in  QemuImg.remove function when image_backend is ceph.

ypu:
@xutian Already update info and check function but just as we talked remove() is another story need support gluster, iscsi and so on. Better to use another pullreq to resolve them all in the same time.

xutian:
@ypu 

that ok for me. currently we support alots of different image backend, the better ways is to solve backup/remove functions in QemuImg class. 

LGTM

thanks,
@xutian

PR #2140, opened 2015-04-13
virttest/virsh: extract common as a function

PandaWei:
Both pool_define_as and pool_create_as needs to check
if the pool type is supported, and these codes are similar.
so extract the common part to reduce code redundancy.


yangdongsheng:
Okey, looks good.

will-Do:
ACK and merge

PR #2141, opened 2015-04-13
virttest/gluster: fix bug for start glusterd

PandaWei:
If glusterd keeps failed status after the first failure,
It would never be started as the orignal code.

In fact, If only the status of glusterd is not active,
we should try to start glusterd again for another cases.

### DIFF
--- a/virttest/gluster.py
+++ b/virttest/gluster.py
@@ -38,7 +38,8 @@ def glusterd_start():
     """
     cmd = "service glusterd status"
     output = utils.system_output(cmd, ignore_status=True)
-    if 'inactive' in output or 'stopped' in output:
+    # The blank before 'active' makes a distinction with 'inactive'
+    if ' active' not in output or 'running' not in output:
         cmd = "service glusterd start"
         error.context("Starting gluster dameon failed")
         output = utils.system_output(cmd)


virttest/gluster.py
```diff
@@ -38,7 +38,7 @@ def glusterd_start():
     """
     cmd = "service glusterd status"
     output = utils.system_output(cmd, ignore_status=True)
-    if 'inactive' in output or 'stopped' in output:
+    if 'active' not in output and 'running' not in output:
```
ypu: Why using and here? Seems or is better for my point of view.

PandaWei: The commit message has mentioned ...
Maybe it's not clear.
Start glusterd maybe failed in some cases, for example because of Selinux.
and it's status will keep 'failed'.
And then, there's another case needs glusterfs and want it's active.
But,
The old code won't to call utils.system_output to perform "service glusterd start".
In fact,
we should try to start gluster again.
This PR can fix this special situation.

will-Do: From your fix code, just one line update about the 'if' condition, so I believe @ypu 's comment is reasonable:
```
if 'active' not in output and 'running' not in output:
-> 
if 'active' not in output or 'running' not in output:

PandaWei: Oops,
I hadn't understand what ypu wants to tell me.
Thanks for explanation.
That's better.
update it now

Hao-Liu:
Hi @PandaWei, 'active' is a sub-string of 'inactive'. I believe it will cause trouble here.

PandaWei:
Yes, you are right.
Add a blank before 'active' to make a distinction with 'inactive'

Hao-Liu:
ACK this and merging. Thanks @PandaWei @ypu.

PR #2143, opened 2015-04-14
virttest/libvirt_vm: check if install package successfully

PandaWei:
Just check the status is not safe,
although the status is zero, but package is not installed.
for example,
the ouput of session.cmd_status_output() as follow,

03:18:59 DEBUG| Sending command: rpm -q qemu-guest-agent ||
yum install -y qemu-guest-agent
03:18:59 DEBUG| Sending command: echo $?
(0, 'rpm -q qemu-guest-agent || yum install -y
qemu-guest-agen<qemu-guest-agent || yum install -y qemu-guest-agent
\x08\0x08\0x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\x08\npackage
qemu-guest-agent is not installed\n
Loaded plugins: product-id, security, subscription-manager\n
This system is not registered to Red Hat Subscription
Management. You can use subscription-manager to register.\n
Setting up Install Process\n
No package qemu-guest-agent available.\nNothing to do\n')

so it's necessary to run 'rpm -q qemu-guest-agent' again on guest.

### DIFF
--- a/virttest/libvirt_vm.py
+++ b/virttest/libvirt_vm.py
@@ -2326,7 +2326,9 @@ class VM(virt_vm.BaseVM):
             # Install the package if it does not exists
             cmd = "rpm -q %s || yum install -y %s" % (name, name)
             status, output = session.cmd_status_output(cmd, timeout=300)
-            if status != 0:
+            # Just check status is not enough
+            # It's necessary to check if install successfully
+            if status != 0 or session.cmd_status("rpm -q %s" % name) != 0:
                 raise virt_vm.VMError("Installation of package %s failed:\n%s" %
                                       (name, output))
         finally:


Hao-Liu:
This code was written assuming yum will return non-zero status if it failed to install a package.
This what I'm getting on RHEL6 and RHEL7:
```
# rpm -q non-exist-pkg || yum -y install non-exist-pkg; echo $?
package non-exist-pkg is not installed
Loaded plugins: langpacks, product-id, subscription-manager
This system is not registered to Red Hat Subscription Management. You can use subscription-manager to register.
Repository epel is listed more than once in the configuration
No package non-exist-pkg available.
Error: Nothing to do
1
```
I'm getting return code 1 for installing an non-exist package with yum. But it seems you get 0 instead in your VM. I'm just curious about how this happened.

PandaWei:
We test it on RHEL5 and expect all of qemu-guest-agent cases skip.
But we encounters failure.

Hao-Liu:
So this could be a problem with old version yum. Then I'm good with this PR. Thanks @PandaWei.

yangdongsheng:
okey, makes sense.

Hao-Liu:
Merging Thanks @PandaWei @yangdongsheng.

PR #2146, opened 2015-04-15
virttest/virsh: return command() directly

PandaWei:
It's no need to store the result of command(cmd, **dargs)
as a variable and then return the variable.


will-Do:
Okay

Hao-Liu:
ACK and merging. Thanks @PandaWei @will-Do.

PR #2151, opened 2015-04-16
utils_misc: simplify codes for operating PCI

PandaWei:
Just compare the return value of utils.run() with
expectation and return True or False directly.
No need to do it by means of if-block statements.


liyangfnst:
OK, simple fix. merge it now

PR #2154, opened 2015-04-16
Fix a bug in rng xml.

rbian:


waynesun09:
ACK and merge the fix.

PR #2156, opened 2015-04-16
Add restart tgtd option in iscsi module

waynesun09:
Add the new param in init and update libvirt iscsi config function to support use it.

### DIFF
--- a/virttest/iscsi.py
+++ b/virttest/iscsi.py
@@ -78,7 +78,7 @@ def iscsi_node_del(target_name=None):
             if target_name in node_tup:
                 cmd = "iscsiadm -m node -o delete -T %s " % target_name
                 cmd += "--portal %s" % node_tup[0]
-                utils.system(cmd)
+                utils.system(cmd, ignore_status=True)
                 break
         if not cmd:
             logging.error("The target '%s' for delete is not in target node"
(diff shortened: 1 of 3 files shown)


CongSmile:
Hi,

On RHEL.7, tgtd service is called targetd.

Could you check it?

Thanks.

waynesun09:
@CongSmile I think you mean "target" service, 'target' and 'tgtd' are two services.

RHEL 7 and upstream is using the LIO kernel target subsystem for iSCSI and other protocols, instead of the tgtd daemon, while tgtd is supported in EPEL on RHEL7.
virt-test did not support LIO now, we still use tgtd as iscsi initiator which with backward compatibility.
LIO support is welcomed and looking forward the progress.

CongSmile:
# service tgtd status
Redirecting to /bin/systemctl status  tgtd.service
tgtd.service
   Loaded: not-found (Reason: No such file or directory)
   Active: inactive (dead)
# uname -r
3.10.0-234.el7.x86_64
# rpm -qa | grep iscsi
iscsi-initiator-utils-6.2.0.873-29.el7.x86_64
libiscsi-1.9.0-6.el7.x86_64
iscsi-initiator-utils-iscsiuio-6.2.0.873-29.el7.x86_64

I'm not sure whether I have missed some packages or others.
But there is no tgtd service on RHEL.7 on my host, it's strange.

Could you check it?

waynesun09:
@CongSmile you need scsi-target-utils rpm

waynesun09:
@CongSmile you need to install EPEL rhel7 repo rpm, then yum install scsi-target-utils rpm
for EPEL:
https://fedoraproject.org/wiki/EPEL

CongSmile:
@waynesun09 

Oh, yes, thanks for you help, I missed the package.

It looks good to me now.

Hao-Liu:
ACK and Merging, Thanks @waynesun09 @CongSmile.

PR #2157, opened 2015-04-16
virt.virttest: update function get_readable_cdroms

xutian:
log debug info when no readable cdrom found


xutian:
ID: 1177445

PR #2161, opened 2015-04-20
small fix in utils_misc.py

PyLearner:
small fix in utils_misc.py

### DIFF
--- a/shared/cfg/guest-os/Linux/Ubuntu/12.04-server.i386.cfg
+++ b/shared/cfg/guest-os/Linux/Ubuntu/12.04-server.i386.cfg
@@ -1,4 +1,5 @@
 - 12.04-server.i386:
+    grub_file = /boot/grub/grub.cfg
     image_name += -12.04-server-32
     vm_arch_name = i386
     os_variant = ubuntuprecise
--- a/virttest/utils_misc.py
+++ b/virttest/utils_misc.py
@@ -500,7 +500,7 @@ def wait_for(func, timeout, first=0.0, step=1.0, text=None):
     :param text: Text to print while waiting, for debug purposes
     """
     start_time = time.time()
-    end_time = time.time() + timeout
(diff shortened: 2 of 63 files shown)


virttest/utils_misc.py
```diff
@@ -500,7 +500,7 @@ def wait_for(func, timeout, first=0.0, step=1.0, text=None):
     :param text: Text to print while waiting, for debug purposes
     """
     start_time = time.time()
-    end_time = time.time() + timeout
+    end_time = time.time() + int(timeout)
```
PandaWei: I really don't see any reason for convert it to int.
Although the timeout is set to int, such as 60, 120, in most cases.
it doesn't mean that we can't set float value to timeout.
the default of step is 1.0. 
so i think float value is supported for timeout

lmr: @PandaWei, correct

lmr: Maybe the author @PyLearner meant to convert to `float` instead of `int`?

lmr:
@PyLearner, you also need to rebase your branch to clean it up and leave only the commits you actually want to contribute with.

PyLearner:
@lmr,@PandaWei  ,sorry to reply late,i made the wrong change ,i will close it ,thanks for your comment!

shared/cfg/guest-os/Linux/Ubuntu/12.04-server.i386.cfg
```diff
@@ -1,4 +1,5 @@
 - 12.04-server.i386:
+    grub_file = /boot/grub/grub.cfg
```
ypu: Please config this in Ubuntu.cfg. As the are all the same so don't need update in separated cfg files.

ypu:
@PyLearner you don't need close it. Just update your pullreq branch will be OK.

PR #2162, opened 2015-04-20
virttest/utils_netperf: fix no return False

PandaWei:
fixed that No return False for function is_target_running.


xutian:
LGTM,

PR #2165, opened 2015-04-21
virttest/utils_libguestfs: No set copyonread explicitly

PandaWei:
copyonread is unavailable on old guestfish and
" copyonread:false" cause errors, such as following on RHEL6.7,

><fs> add-drive-opts /iso/g/images/rhel6X-64.qcow2 \
      readonly:true copyonread:false
add-drive-opts: unknown optional argument "copyonread:false"
><fs>

And the default is false for copyonread,
so it's no need to set " copyonread:false" explicitly.
Please refer to manual for more details.

### DIFF
--- a/virttest/utils_libguestfs.py
+++ b/virttest/utils_libguestfs.py
@@ -531,7 +531,10 @@ class GuestfishPersistent(Guestfish):
         if copyonread:
             cmd += " copyonread:true"
         else:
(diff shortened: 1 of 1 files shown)


will-Do:
ACK

Hao-Liu:
ACK and merging. Thanks @PandaWei @will-Do

PR #2166, opened 2015-04-21
Support to -nodefconfig into QEMU command line.

pvital:
The -nodefconfig parameter prevents QEMU from loading the
sysconfdir/qemu.conf and sysconfdir/target-ARCH.conf files at startup.

This patch adds support to virt-test add this parameter if specified by
user while launching virt-test.

Signed-off-by: Paulo Vital <paulo.vital@profitbricks.com>


lmr:
OK, I've tested `./run -t qemu` and `./run -t qemu --defconfig no` and both produced expected outcomes. This looks good to me, approved, thanks!

PR #2168, opened 2015-04-23
virttest/libvirt_storage: elaborate on pool_delete

PandaWei:
The function pool_delete() doesn't call virsh sub-command
pool-delete to delete a pool,
which looks without rhyme or reason.
This patch make some explanation for it.


will-Do:
ACK

waynesun09:
ACK and merging now, Thanks.

PR #2169, opened 2015-04-24
virttest/qemu_vm: fix bug related to try-except-else

PandaWei:
If IndexError is raised,
The else-block statements never be run at graceful_shutdown().

Explanation As following,
try:
     ...
except (IndexError), e:
     try:
            session = self.serial_login()
     except (remote.LoginError, virt_vm.VMError), e:
            logging.debug(e)
     >>>Although successfully get session by serial_login()
     >>>The following else-block never be run.
except (remote.LoginError, virt_vm.VMError), e:
     ...
else:
      >>>send shutdown command to VM if no error occurs.

This patch is dedicated to fix it.

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -2578,6 +2578,14 @@ class VM(virt_vm.BaseVM):
         True. If QEMU is running in -no-shutdown mode, the QEMU process
         may be still alive.
         """
+        def _shutdown_by_sendline():
+            try:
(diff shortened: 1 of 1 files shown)


will-Do:
Good catch, ack this fix

waynesun09:
ACK

Hao-Liu:
Merging this. Thanks @PandaWei @will-Do @waynesun09.

PR #2177, opened 2015-05-05
virttest.shared.unattended: Remove dhclient command from SLES 12

ypu:
...ES 12

The old command will only update the line from no to yes. But in SLES 12 it still
start with '#' and the login in following cases will failed.

ID: 1218472

### DIFF
--- a/shared/unattended/SLES-12.xml
+++ b/shared/unattended/SLES-12.xml
@@ -549,8 +549,7 @@ echo ttyS0 >> /etc/securetty]]>
       <script>
         <debug config:type="boolean">true</debug>
         <filename>config</filename>
-        <source><![CDATA[dhclient eth0
-chkconfig sshd on
+        <source><![CDATA[chkconfig sshd on
 sed -i -e 's/\(PasswordAuthentication\s\)no/\1yes/g'  /etc/ssh/sshd_config
 service sshd restart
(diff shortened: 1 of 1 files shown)


shared/unattended/SLES-12.xml
```diff
@@ -551,7 +551,7 @@ echo ttyS0 >> /etc/securetty]]>
         <filename>config</filename>
         <source><![CDATA[dhclient eth0
 chkconfig sshd on
-sed -i -e 's/\(PasswordAuthentication\s\)no/\1yes/g'  /etc/ssh/sshd_config
+sed -i -e 's/*PasswordAuthentication.*/PasswordAuthentication yes/g'  /etc/ssh/sshd_config
```
xiagao: hi, ypu

Maybe this command can't change the file.

ypu: @xiagao Thanks for your commnet. Seems this line is not the root cause of the script failed to update the ssh cfg. Will do more test with it and update the pullreq later.

FengYang:
I cannot open the id, but the patch looks good to me. : )

Acked-by: FengYang

xutian:
LGTM,

ypu:
Thanks @FengYang for your reviewing. :)

PR #2178, opened 2015-05-11
Two Commits for env_process.py

PandaWei:
1)  virttest/env_process: remove unused variables
    
    The codes used to start tcpdump had been rewritten
    into a new function called _start_tcpdump at utils_env.py.
    But, It didn't discard unused variables by 4b518ed.

2)    virttest/env_process: fix exception string
    
    If the exception has been logged by logging.error("e"),
    Only one letter e will be output.
    so replace "e" with str(e).


Hao-Liu:
Ack and merge for trivial fix. Thanks @PandaWei

PR #2180, opened 2015-05-11
utils_test/__init__.py: rewrite get_free_space()

PandaWei:
The codes of get_free_space call self.runner.run twice,
and No catch IndexError/ValueError for the following,
    int(line.split()[3]) / 1048576
Besides above, It's better to output the return of
self.runner.run for failure case, not for successful case.
and so on.

So need to rewirte and make it more better.

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -1800,28 +1800,39 @@ class RemoteDiskManager(object):
     def get_free_space(self, disk_type, path='/', vgname=None):
         """
         Get free space of remote host for path.
+
+        :return : the unit is 'G'.
         """
         if disk_type == "file":
-            directory = os.path.dirname(path)
-            try:
-                output = self.runner.run("df %s" % directory).stdout
-                logging.debug(output)
-            except error.CmdError, detail:
-                raise error.TestError("Get %s space failed:%s" % (directory,
-                                                                  detail))
-            for line in output.splitlines()[1:]:
-                g_size = int(line.split()[3]) / 1048576
-                return g_size
-            raise error.TestError("Get %s space failed." % directory)
+            cmd = "df -BG %s" % os.path.dirname(path)
         elif disk_type == "lvm":
-            output = self.runner.run("vgs --units=g | grep %s" % vgname).stdout
-            if re.search(vgname, output.stdout):
+            cmd = "vgs --units=g | grep %s" % vgname
+        else:
+            raise error.TestError("Unsupported Disk Type %s" % disk_type)
+
+        try:
+            output = self.runner.run(cmd).stdout
+        except error.CmdError, detail:
+            logging.debug(output)
+            raise error.TestError("Get space failed: %s." % str(detail))
+
+        if disk_type == "file":
+            try:
+                return int(output.splitlines()[1].split()[3].split('G')[0])
+            except IndexError, detail:
+                raise error.TestError("Get %s space failed: %s." %
+                                      (os.path.dirname(path), str(detail)))
+        elif disk_type == "lvm":
+            if re.search(vgname, output):
                 try:
-                    return int(output.split('g')[0])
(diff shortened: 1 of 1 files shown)


virttest/utils_test/__init__.py
```diff
@@ -1802,26 +1802,35 @@ def get_free_space(self, disk_type, path='/', vgname=None):
...
                 except (IndexError, ValueError), detail:
-                    output = detail
-            raise error.TestError("Get VG %s space failed:%s" % (vgname,
-                                                                 output))
+                    raise error.TestError("Get %s space failed: %s." %
+                                          (vgname, str(detail)))
+            else:
+                raise error.TestError("Get %s space failed: %s." %
+                                      (vgname, output))
+        return avail_size
```
yangdongsheng: What's the unit of the ret-value.

PandaWei: It's 'G'

yangdongsheng: Okey, maybe a comment for this function would help a lot. And then you can use "df -BG" to get the result in GiB unit. In addition, is there any bug in the original implementaion? Or that's just a cleanup again?

PandaWei: yes, it's just clean up.
I will update it as your comments.

PandaWei:
1) fix bug(s) for 'lvm' type when match free space
   - call .stdout twice that leads to no attribute 'stdout'
     output = self.runner.run("vgs --units=g | grep %s" % vgname).stdout
     if re.search(vgname, output.stdout):
     (Pdb) output.stdout
     *** AttributeError: 'str' object has no attribute 'stdout'

   - Old expression doesn't get correct space number, as following,
     [root@localhost tp-libvirt]# vgs --units=g
      VG     #PV #LV #SN Attr   VSize  VFree
     fedora   1   2   0 wz--n- 49.15g    0g
     [root@localhost tp-libvirt]# python
     >>> output="  fedora   1   2   0 wz--n- 49.15g    0g"
     >>> output.split('g')[0]
     '  fedora   1   2   0 wz--n- 49.15'
2) Using option '-BG' to get space directly by G.
   Old code doesn't catch IndexError for 'int(line.split()[3])/1048576'.
3) add explanation for the unit of return value.
4) cleanup codes and avoid call twice for self.runner.run().

virttest/utils_test/__init__.py
```diff
@@ -1800,28 +1800,39 @@ def __init__(self, params):
...
-                g_size = int(line.split()[3]) / 1048576
-                return g_size
-            raise error.TestError("Get %s space failed." % directory)
+                return int(output.splitlines()[1].split()[3].split('G')[0])
+            except IndexError, detail:
+                raise error.TestError("Get %s space failed: %s." %
+                                      (os.path.dirname(path), str(detail)))
         elif disk_type == "lvm":
-            output = self.runner.run("vgs --units=g | grep %s" % vgname).stdout
-            if re.search(vgname, output.stdout):
```
yangdongsheng: That's really a bug, because the output is a str rather than a Result object, then there is no .stdout in output. Good catch!!.

yangdongsheng:
Thanx for your summary of your patch, that would help a lot to reviewer. And your patch looks much better in this version. Good job.

yangdongsheng:
If you past the result of the cases which are affected by this patch, I think we can let it go.  I believe there is a case which failed would pass after this commit.

PandaWei:
@yangdongsheng
Thanks for your comments，
In fact, I haven't no case failed by that problem.
Initially, I just want to cleanup code and pay close attention to file type, not lvm type.

You asked me to do more tests about my patch.
so i wrote some small scripts to call this function and verify it.
in this process, i found the problem of lvm.

so thanks to your reminder， i found this bug。

will-Do:
Okay, good fix.
Now I think we can merge this patch.

PR #2182, opened 2015-05-11
virsh.py: Bigger timeout is needed on slower connection/systems

svirt:
Bigger maximum timeout is needed on slower connection/systems

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -174,7 +174,7 @@ class VirshSession(aexpect.ShellSession):
         if ssh_remote_auth or self.uri:
             # Handle ssh / password prompts
             remote.handle_prompts(self, self.remote_user, self.remote_pwd,
-                                  prompt, debug=True)
+                                  prompt, timeout=60, debug=True)
 
         # fail if libvirtd is not running
         if self.cmd_status('list', timeout=60) != 0:


PandaWei:
Hi,
I don't think it's a good idea to modify the default of function parameter.
Not all of case needs longer timeout.
You can set any maximum timeout by passing to this function when call it.

svirt:
Hi,
Yes, that might be a better idea. I will check and modify the call function.
thanks

svirt:
**@PandaWei**
moved timeout value to the function call for the virsh.
thanks a lot for the review

PandaWei:
Do you sure 20s is enough and verify it?

svirt:
I ran a few times the tests in one of my quite slow environments, it is enough there. 
As with default value 10  in the mornings I encounter failures with timeout errors for nightly test runs (randomly usually)

PandaWei:
As you said, 20s is enough,

And as codes, it will last more 30 sec if encounter timeout and hint ""Please wait"
           elif match == 7:  # "Please wait"
                if debug:
                    logging.debug("Got 'Please wait'")
                timeout = 30

In fact,
It cost 20+30, which close to 1 min.
so, i think set it to 60sec more reasonable.

Do you think so?

svirt:
I think that also could be a reasonable point. Probably in the particular case you mentioned 60 seconds period is a better timeout.
From my testing perspective 60 seconds would not harm. Shell I change it to 60 seconds and recommit?
thanks

PandaWei:
Now, no need to recommit.
let's listen to others.
Maybe other people have different opinion.

svirt:
Ok, I am flexible to change it whenever it's needed.
thanks

svirt:
I guess 60 seconds is good value , did some tests , was ok

PandaWei:
Ack

svirt:
thanks, i will now recommit the 60 seconds version

svirt:
Hi, thanks for the review.

similar issue as with https://github.com/autotest/tp-libvirt/pull/554

I think probably so far we can delay with those timeout changes application, as seemed there was a network/ssh related problem causing timeouts and even on slower system the current settings are fine after fixing those problems. I will see how are going the tests in the future and maybe if needed will come with those timeout changes.

will-Do:
@svirt 
Any update about this PR? If the fix is satisfy your test case, we can merge it.

svirt:
@will-Do 
Hi, I think for now it's better to close/cancel this merge, as my many experiments showed that the root cause of the delays usually is network related and the incremented time does not help in all the cases. Do you agree if I close/cancel this merge? Thanks

will-Do:
@svirt 
Sure, I will close it directly.

svirt:
thank you

PR #2199, opened 2015-05-27
virt: enable dump guest core

ShaolongHu:
Signed-off-by: Shaolong Hu <shu@redhat.com>

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -892,6 +892,12 @@ class DevContainer(object):
             if machine_type in m_types:
                 if (self.has_option('M') or self.has_option('machine')):
                     cmd = "-M %s" % machine_type
+                    # default yes in qemu, only deal with 'no'
+                    if params['dump_guest_core'] == 'no':
+                        # new in 6.7 and 7.2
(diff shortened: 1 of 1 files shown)


virttest/qemu_devices/qcontainer.py
```diff
@@ -892,6 +892,12 @@ def machine_other(cmd=False):
             if machine_type in m_types:
                 if (self.has_option('M') or self.has_option('machine')):
                     cmd = "-M %s" % machine_type
+                    # default yes in qemu, only deal with 'no'
+                    if params['dump_guest_core'] == 'no':
```
ypu: Please use params.get() here. params[''] will raise error if we don't have related parameter setup in cfg.

virttest/qemu_devices/qcontainer.py
```diff
@@ -892,6 +892,12 @@ def machine_other(cmd=False):
...
+                        # new in 6.7 and 7.2
+                        if ('7.2' in params['host_kernel_ver_str'] or
+                                '6.7' in params['host_kernel_ver_str']):
```
ypu: We don't need this, we have our own cfg files to control host version

PR #2200, opened 2015-05-29
virt: enable machine_type accepts extra parameters

ShaolongHu:
Some new feature like dump-guest-core in qemu asks for command line
like '-machine pc,dump-guest-core=on/off', so we need let machine
type accepts extra parameters, and "-M" can not accept extra parameter
in qemu side, use "-machine".

ID: 1226116

Signed-off-by: Shaolong Hu <shu@redhat.com>

### DIFF
--- a/virttest/qemu_devices/qcontainer.py
+++ b/virttest/qemu_devices/qcontainer.py
@@ -884,6 +884,7 @@ class DevContainer(object):
             return devices
 
         machine_type = params.get('machine_type')
+        machine_type_extra_params = params.get('machine_type_extra_params')
         if machine_type:
             m_types = []
             for _ in self.__machine_types.splitlines()[1:]:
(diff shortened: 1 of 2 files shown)


virttest/qemu_devices/qcontainer.py
```diff
@@ -891,7 +892,9 @@ def machine_other(cmd=False):
 
             if machine_type in m_types:
                 if (self.has_option('M') or self.has_option('machine')):
-                    cmd = "-M %s" % machine_type
+                    cmd = "-machine %s" % machine_type
+                    if machine_type_extra_params:
+                        cmd = "%s,%s" % (cmd, machine_type_extra_params)
```
ypu: 1. cmd += ",%s" % machine_type_extra_params should work for this condition.
2. this only in machine_type exist branch but not in the auto judgement branch.
3. at lease need strip machine_type_extra_params wiht "," to ignore some problems caused be cfg file.

ShaolongHu:
@ypu 

Has updated:

There is chance that "-M" and "-machine" both are absent, but
machine_type_extra_params exists, in this case we apply extra params
to default machine type, to be noted, qemu won't accept command like
"-machine ,xxx=on", there must be specific machine type, so we use
"q35"/"pc" for Q35/i440FX respectively. I leave aarch64 and isa/unknown
machine type ignore machine type extra params currently, for i am not
sure the appropriate behavior for them.

ypu:
LGTM

spcui:
Acked-by: @spcui

ypu:
Thanks @ShaolongHu amd @spcui will merge it.

PR #1924, opened 2014-12-18
Network work related function update

FengYang:
ID: 1173392 1173393 1172957

PR #2201, opened 2015-05-29
virttest.utils_test: Update ping func for windows guest

spcui:
Not supported "-c" option using ping command in windows
guest, should be used "-n count" instead it.

Signed-off-by: Shuping Cui <scui@redhat.com>

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -1213,7 +1213,12 @@ def ping(dest=None, count=None, interval=None, interface=None,
     else:
         command += " localhost "
     if count is not None:
-        command += " -c %s" % count
+        count_str = " -c %s" % count
+        if session:
+            output = session.cmd_output("ping")
+            if "-n count" in output:
+                count_str = " -n %s" % count
+        command += count_str
     if interval is not None:
         command += " -i %s" % interval
     if interface is not None:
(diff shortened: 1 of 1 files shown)


spcui:
ID: 1214553

virttest/utils_test/__init__.py
```diff
@@ -1213,7 +1213,12 @@ def ping(dest=None, count=None, interval=None, interface=None,
     else:
         command += " localhost "
     if count is not None:
-        command += " -c %s" % count
+        count_str = " -c %s" % count
+        if session:
+            output = session.cmd_output("ping")
```
ypu: @spcui this will raise cmd error in the code. And don't think this is a good idea to judge the os type.

virttest/utils_test/__init__.py
```diff
@@ -1213,7 +1213,12 @@ def ping(dest=None, count=None, interval=None, interface=None,
...
+            if "-n count" in output:
+                count_str = " -n %s" % count
+        command += "%s" % count_str
```
ypu: sorry for the wrong comments. as the cmd_output will not check status now. So above check look good to me.

But seems here need update. command += count_stre will work.

spcui:
@ypu updated with your comment, please review again, thanks.

ypu:
Hi @spcui checked the function again. Seems the flood ping part also need update. There are some Linux special commands like "kill" hard code in our code. Can you update it in the same time?

spcui:
Closed it, because this pullreq (https://github.com/autotest/virt-test/pull/1924) already fixed this issue.

PR #2204, opened 2015-06-02
utils_test.libvirt: Add iscsi volume disk type

waynesun09:
Add iscsi volume disk type in set_vm_disk, iscsi volume type require
define iscsi pool first for using the volume in vm.

Check more disk info in:
http://libvirt.org/formatdomain.html#elementsDisks


waynesun09:
Merge this as related to iscsi change

PR #2214, opened 2015-06-08
shared.cfg.machines: update pseries config

luckyh:
ID: 1228581


ypu:
@luckyh thanks for your update and can you also add ppcle guest cfg in upstream too? Seems upstream still don't have it. So it is may a little confuse to see only ppc64le here.

ypu:
LGTM

ypu:
Thanks @luckyh will merge it.

PR #2221, opened 2015-06-17
Change the virsh blkdeviotune's options processing method.

liuzzfnst:
The old method for options in virsh blkdeviotune just fits
single option, but '--live --config' should be given at the same time.


PandaWei:
LGTM, 
Indeed The PR is useful for the case that using two or more than two options

will-Do:
Ok, apply this patch

PR #2222, opened 2015-06-17
Add config and snapshot tag to disk xml.

rbian:


waynesun09:
ACK and merge now, thanks @rbian

PR #2223, opened 2015-06-17
Modify for pool problem fix

kylazhang:
1. dir create in pre_pool sometimes duplicated with pool_build which may case error in new version,
    so add 1 more condition for dir creat
2. before list volume, must make sure pool is active, else it will report error


will-Do:
ACK

waynesun09:
ACK and merging now, thx @kylazhang @will-Do

PR #2230, opened 2015-06-26
virttest.qemu_storage: Support "-o options" for qemu_img.convert.

Chenditang:


virttest/qemu_storage.py
```diff
@@ -198,8 +202,19 @@ def convert(self, params, root_dir, cache_mode=None):
         cmd += " convert"
         if convert_compressed == "yes":
             cmd += " -c"
+        cmd += " -o "
```
xutian: Hi @Chenditang 

options = [ ]
if xxx:
       options.append("attr=val")
...
if options:
       qemu_img_cmd += " -o %s" % ",".join(options)

looks more compact.

xutian:
LGTM,

will-Do:
ACK

PR #2232, opened 2015-07-01
Rbd disk type support.

rbian:
libvirt_storage, virsh: add rbd pool support; modify pool_define_as function to make target as the optional
disk, pool_xml, secret_xml: add some new rbd pool/volume attributes to xml.

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -1819,7 +1819,7 @@ def _pool_type_check(pool_type):
     :param pool_type: pool type
     :return: valid pool type or None
     """
-    valid_types = ['dir', 'fs', 'netfs', 'disk', 'iscsi', 'logical', 'gluster']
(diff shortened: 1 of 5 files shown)


virttest/virsh.py
```diff
@@ -2009,7 +2009,7 @@ def pool_state_dict(only_names=False, **dargs):
     return result
 
 
-def pool_define_as(name, pool_type, target, extra="", **dargs):
+def pool_define_as(name, pool_type, target="", extra="", **dargs):
     """
```
will-Do: I saw the func doc mentioned pool type, so you can also update it:
1). typo: s/typ/type
2). Add 'rbd' type desc

rbian: Thanks @will-Do , I have updated the patch.

will-Do:
ACK

waynesun09:
ACK

PR #2234, opened 2015-07-02
virt-test: add check before add virtocosole device

xutian:


xutian:
ID: 1228538

xutian:
@spcui , no need to do it,  we will keep internal code to run with  python <= python2.4 and old qemu version.

xutian:
@CongSmile  and @spcui 

Any comment on this change?

spcui:
OK, LGTM
Acked-by: scui@redhat.com

PR #2238, opened 2015-07-05
virt-test: test_setup.py: Add class for enable multi queue in guest.

CongSmile:
Enable multi queue for those guest whose os are 7.* and 6.6+.

Signed-off-by: Cong Li <coli@redhat.com>

### DIFF
--- a/virttest/env_process.py
+++ b/virttest/env_process.py
@@ -805,6 +805,17 @@ def preprocess(test, params, env):
     # Preprocess all VMs and images
     if params.get("not_preprocess", "no") == "no":
         process(test, params, env, preprocess_image, preprocess_vm)
+        # Enable multi queue in guest
+        mq = test_setup.MultiQueueConfig(params)
+        if params.get("nic_model") == "virtio":
+            if params.get("setup_mq") == "yes" and params.get("queues"):
+                queues = int(params.get("queues", 1))
+                if queues == 1:
+                    logging.info("No need enable MQ feature for single queue")
+                elif queues > 1:
+                    for vm_name in params.get("vms").split():
+                        vm = env.get_vm(vm_name)
+                        mq.setup(queues, vm_name, vm)
 
     # Start the screendump thread
     if params.get("take_regular_screendumps") == "yes":
--- a/virttest/test_setup.py
+++ b/virttest/test_setup.py
@@ -14,6 +14,7 @@ import data_dir
 import utils_misc
 import versionable_class
 import openvswitch
+from virttest import utils_net
 
 try:
     from virttest.staging import utils_memory
@@ -1754,3 +1755,62 @@ class EGDConfig(object):
             utils.wait_for(_all_killed, timeout=60)
         except OSError:
             logging.warn("egd.pl is running")
+
+
+class MultiQueueSetupError(Exception):
+
+    """
+    Raise when setup multi queue failed in guest.
+    """
+    pass
+
+
+class MultiQueueConfig(object):
+
+    """
+    Enable multi queue for virtio nics in guest if queues > 1.
+    """
+
+    def __init__(self, params):
+        """
+        :param params: Dict like object containing parameters for the test.
+        """
+        self.params = params
+
+    def enable_mq(self, queues, vm_name, vm):
+        """
+        Enable multi queue in guest
+        """
+        error.context("Login to guest\n")
+        session = vm.wait_for_login()
+        for i, nic in enumerate(vm.virtnet):
+            if "virtio" in nic['nic_model']:
+                ifname = utils_net.get_linux_ifname(session,
+                                                    vm.get_mac_address(i))
+                enable_mq_cmd = "ethtool -L %s combined %s" % (ifname, queues)
+                mq_show_cmd = "ethtool -l %s" % ifname
+                logging.debug("Enable multi queue for nic %s in %s:"
+                              % (ifname, vm_name))
+                show_nic_info = session.cmd(mq_show_cmd)
+                logging.info("Before MultiQueue Changed of nic %s is %s"
+                             % (ifname, show_nic_info))
+                try:
+                    if enable_mq_cmd:
+                        output = session.cmd_output(enable_mq_cmd, timeout=300)
+                except Exception:
+                    msg = "Enable multi queue in %s failed" % vm_name
+                    msg += "of nic" % ifname
+                    msg += "The error info is: %s" % output
+                    raise MultiQueueSetupError(msg)
+                show_nic_info = session.cmd(mq_show_cmd)
+                logging.info("After MultiQueue Changed of nic %s is %s"
+                             % (ifname, show_nic_info))
+        session.close()
+
+    def setup(self, queues, vm_name, vm):
+        """
+        :param queues: int, the number of queues.
+        :param vm_name: string, vm's name, especially for the multi vms.
+        :param vm: VM object.
+        """
+        self.enable_mq(queues, vm_name, vm)
--- a/shared/cfg/guest-os/Linux/RHEL/6.6.cfg
+++ b/shared/cfg/guest-os/Linux/RHEL/6.6.cfg
@@ -2,6 +2,7 @@
     no setup
     image_name = images/rhel66
     os_variant = rhel6
+    setup_mq = yes
     unattended_install, check_block_size.4096_512, check_block_size.512_512, svirt_install:
         unattended_file = unattended/RHEL-6-series.ks
         syslog_server_proto = udp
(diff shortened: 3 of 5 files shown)


CongSmile:
id: 1203966

virttest/env_process.py
```diff
@@ -805,6 +805,11 @@ def preprocess(test, params, env):
     # Preprocess all VMs and images
     if params.get("not_preprocess", "no") == "no":
         process(test, params, env, preprocess_image, preprocess_vm)
+        mq = test_setup.MultiQueueConfig(params)
+        if params.get("setup_mq") == "yes" and params.get("queues"):
```
xutian: I think here should be "virtio" in params.get("nic_model") and setup_mq == yes ?

virttest/env_process.py
```diff
@@ -805,6 +805,11 @@ def preprocess(test, params, env):
...
+            queues = int(params.get("queues"))
+            if queues > 1:
```
xutian: if queues < 1, qemu should refuse to start else it should be a bug. so no need to check it here.

virttest/test_setup.py
```diff
@@ -1754,3 +1754,47 @@ def _all_killed():
             utils.wait_for(_all_killed, timeout=60)
         except OSError:
             logging.warn("egd.pl is running")
+
+
+class MultiQueueConfig(object):
```
xutian: doc string ???

virttest/test_setup.py
```diff
@@ -1754,3 +1754,47 @@ def _all_killed():
...
+
+    def __init__(self, params):
+        self.params = params
+
+    def find_nic(self, session):
+        show_nic_cmd = "ls /sys/class/net"
+        nics = []
```
xutian: no need here..

virttest/test_setup.py
```diff
@@ -1754,3 +1754,47 @@ def _all_killed():
...
+        nics = session.cmd_output(show_nic_cmd).split(" ")
+        for nic in nics:
+            find_nic_cmd = "ethtool -i %s | grep virtio" % nic
```
xutian: better to  filter key words  in next sentence, so that we can see all nics in log whether test failed or not.

virttest/test_setup.py
```diff
@@ -1754,3 +1754,47 @@ def _all_killed():
...
+            status, output = session.cmd_status_output(find_nic_cmd, timeout=300)
+            if status != 0:
+                raise error.TestFail("No virtio nic, can't enable multiqueue\n")
+            else:
+                match = str(nic)
+                return match
```
xutian: if we have many virtio nics, you current logical will only set mq for first match one only, so ??

virttest/test_setup.py
```diff
@@ -1754,3 +1754,47 @@ def _all_killed():
...
+            logging.debug("Enable multi queue in guest\n")
+            mq_show_cmd = "ethtool -l %s" % nic
+            logging.debug("Before MultiQueue Changed:\n")
+            session.cmd(mq_show_cmd)
+            try:
+                status, output = session.cmd_status_output(enable_mq_cmd, timeout=300)
+                if status != 0:
+                    raise error.TestFail("Enable multi queue in guest failed!\n"
+                                         "The error info is: %s\n" % output)
+                else:
```
xutian: only record it when failed is enough.

virttest/env_process.py
CongSmile: Yes, have updated in the code.

virttest/env_process.py
CongSmile: I just want to judge that if queues==1, no need to enable multi queue

virttest/test_setup.py
CongSmile: update it in the code.

virttest/test_setup.py
CongSmile: Yes, it's a better way, updated.

virttest/test_setup.py
CongSmile: updated.

virttest/test_setup.py
CongSmile: updated.

CongSmile:
Hi @xutian 

I have updated the code, fix the problem you mentioned and it fit for multi vms and multi nics now.

Could you help review it again?

Thanks a lot.

virttest/test_setup.py
```diff
@@ -1754,3 +1754,66 @@ def _all_killed():
...
+        nics = session.cmd_output(show_nic_cmd).split()
+        match = []
+        for nic in nics:
+            get_nic_info_cmd = "ethtool -i %s" % nic
+            output = session.cmd_output(get_nic_info_cmd, timeout=300)
+            if "virtio" in output:
+                match.append(str(nic))
+            else:
+                continue
+        if match == []:
```
xutian: if not match:

virttest/test_setup.py
```diff
@@ -1754,3 +1754,66 @@ def _all_killed():
...
+            raise error.TestFail("No virtio nic, can't enable multiqueue\n")
```
xutian: do we need raise error here ?

virttest/test_setup.py
```diff
@@ -1754,3 +1754,66 @@ def _all_killed():
...
+                              % (nic, vm_name))
+                mq_show_cmd = "ethtool -l %s" % nic
+                logging.debug("Before MultiQueue Changed of nic %s:" % nic)
+                session.cmd(mq_show_cmd)
+                status, output = session.cmd_status_output(enable_mq_cmd, timeout=300)
+                if status != 0:
+                    raise error.TestFail("Enable multi queue in %s failed"
+                                         "of nic %s! The error info is: %s"
+                                         % (vm_name, nic, output))
+                logging.debug("After MultiQueue Changed:")
```
xutian: what kinds of info do you want to let test know?

virttest/test_setup.py
```diff
@@ -1754,3 +1754,66 @@ def _all_killed():
...
+            enable_mq_cmd = "ethtool -L %s combined %s" % (nic, queues)
+            if enable_mq_cmd:
+                logging.debug("Enable multi queue for nic %s in %s:"
+                              % (nic, vm_name))
+                mq_show_cmd = "ethtool -l %s" % nic
+                logging.debug("Before MultiQueue Changed of nic %s:" % nic)
+                session.cmd(mq_show_cmd)
+                status, output = session.cmd_status_output(enable_mq_cmd, timeout=300)
+                if status != 0:
+                    raise error.TestFail("Enable multi queue in %s failed"
```
xutian: It's better to define exception like MutilQueueSetupError  to told test what happen when setup mutil-queues

virttest/test_setup.py
CongSmile: Yes, I think it is a need.

We have do the judgement  in env_process.py. 
If the code goes there, it must needs enable mq in guest;
else it should be a bug I think.
+        if params.get("nic_model") == "virtio":
+            if params.get("setup_mq") == "yes" and params.get("queues"):

virttest/test_setup.py
```diff
@@ -1754,3 +1754,66 @@ def _all_killed():
...
+        error.context("Login to guest\n")
+        session = vm.wait_for_login()
+        nics = self.get_virtio_nic(session)
+        for nic in nics:
+            enable_mq_cmd = "ethtool -L %s combined %s" % (nic, queues)
+            if enable_mq_cmd:
+                logging.debug("Enable multi queue for nic %s in %s:"
+                              % (nic, vm_name))
+                mq_show_cmd = "ethtool -l %s" % nic
+                logging.debug("Before MultiQueue Changed of nic %s:" % nic)
```
xutian: if you want to show info about nic, please get it then print logging info.

virttest/test_setup.py
CongSmile: I want to show the queues info in guest before and after changing the queues,
which can give a comparision of the queues.

virttest/test_setup.py
CongSmile: Ok, that would be better, will update it.

Thanks.

virttest/test_setup.py
xutian: We have default values for queues for virtio-nic,  by default queues should be enable in guest by default(guest virtio-nic driver should do it, if not I think it's a bug device driver).  In some user case,  user want to change queues via tool like ethool in linux guest is allow.
 So if this PR is implement change queues in guest I think that's acceptable but if use to enable queues before test , I think it's not acceptable.

Thanks
@xutian

xutian:
Nacked  this PR, I think this actions is really not acceptable for test.

Thanks,
@xutian

lmr:
Hi @CongSmile. As you might know, we're moving virt-test to avocado-vt, so I please ask you to consider re-implementing this functionality in the later. The new repo is:

Please keep in mind the following guidelines when creating the new PR:

1) Most of the time you can pick the patches of your original PR and merge them on a new branch with minimal changes.
3) Don't use autotest APIs, since we're moving away from using them. Please review the code and change things like `utils.system` -> `process.system`, `utils.run` -> `process.run`, so on and so forth.

Thanks, and I'm sorry for any inconveniences we might have caused your team.

PR #2240, opened 2015-07-06
Guest login to desktop instead of start screen after guest login for w…

vipmike007:
…indows8 and windows 8.1

Fix issue guest will come into blank screen after no response in 10 mins
ID: 1146402


vipmike007:
~spcui , pls check again

PR #2241, opened 2015-07-08
virttest.qemu_storage: Support "--backing-chain" for qemu_img.info

Chenditang:


virttest/qemu_storage.py
```diff
@@ -343,16 +343,21 @@ def remove(self):
         else:
             logging.debug("Image file %s not found", self.image_filename)
 
-    def info(self):
+    def info(self, params):
```
xutian: Hi,

We need to keep our API stably, so default value is must at here.

virttest/qemu_storage.py
```diff
@@ -343,16 +343,21 @@ def remove(self):
...
         """
         Run qemu-img info command on image file and return its output.
```
xutian: Update docstring please !!!

virttest/qemu_storage.py
```diff
@@ -343,16 +343,21 @@ def remove(self):
...
         """
         logging.debug("Run qemu-img info comamnd on %s", self.image_filename)
+
+        backing_chain = params.get("backing_chain")
```
xutian: if we set backing_chain = no in  config file, looks --backing_chain will be used too.  this test is not good enough

Chenditang:
@xutian 
Updated code as your suggestion.
Thank you very much~

virttest/qemu_storage.py
```diff
@@ -343,16 +343,23 @@ def remove(self):
         else:
             logging.debug("Image file %s not found", self.image_filename)
 
-    def info(self):
+    def info(self, params={}):
```
will-Do: You should not use a mutable object as the default value. For the detail, pls google it.

Chenditang:
@will-Do @xutian 
Updated, delete the params parameter.

Thanks!

virttest/qemu_storage.py
```diff
@@ -348,11 +348,15 @@ def info(self):
         Run qemu-img info command on image file and return its output.
         """
         logging.debug("Run qemu-img info comamnd on %s", self.image_filename)
+        backing_chain = self.params.get("backing_chain")
         cmd = self.image_cmd
+        cmd += " info"
+        if backing_chain == "yes":
+           cmd += " --backing-chain"
```
xutian: we need to check is qemu-img support this options first, so that we can ensure it not block  our test on z-stream and rhel6 host.

PR #2244, opened 2015-07-09
Reinstall guest agent when existing one is masked

kylazhang:


kylazhang:
Any one please help to review it asap as it affect out qe consumption test, thanks

virttest/libvirt_vm.py
```diff
@@ -2393,6 +2393,11 @@ def _start_ga():
             if not _is_ga_running():
                 cmd = "service qemu-guest-agent start"
                 status, output = session.cmd_status_output(cmd)
+                # Sometimes agent serverice may be masked so need to reinstall
```
PandaWei: Please fix typo

PandaWei:
Is there any failed case caused by the problem?

PandaWei:
Please attach the details, such as error message..

kylazhang:
@PandaWei Sometimes it may report error

  ERROR| VMError: Start qemu-guest-agent failed:
  ERROR| Redirecting to /bin/systemctl start  qemu-guest-agent.service
  ERROR| Failed to issue method call: Unit qemu-guest-agent.service is masked.

The resolve method is to reinstall the package

virttest/libvirt_vm.py
kylazhang: ok. thanks

PandaWei:
Thanks for you reply.
But fix the masked problem by reinstall is a bad idea，
which is time-consuming operation and unnecessary.
Please consider using 'systemctl unmask  service'  to fix it.
I‘m not sure, please have a try at least.

kylazhang:
@PandaWei Thanks for your advice, I try and it does not work. Any other methods?

PandaWei:
I have no suggestion so for.
Let's listen to the view of others.

will-Do:
This should be BZ#https://bugzilla.redhat.com/show_bug.cgi?id=1188148 , as file qemu-guest-agent.service is broken(link to /dev/null or file corrupted), then you can rebuild the service link or reinstall it.
I'm okay with this patch, but you can add a more precise desc in the code

kylazhang:
The desc has been changed. Thanks

will-Do:
Thx

PR #2246, opened 2015-07-10
virttest.qemu_vm: Add function for disable-legacy and disable-modern

spcui:
Signed-off-by: Shuping Cui <scui@redhat.com>

### DIFF
--- a/virttest/qemu_vm.py
+++ b/virttest/qemu_vm.py
@@ -1149,6 +1149,24 @@ class VM(virt_vm.BaseVM):
                 dev.set_param("id", devid)
             devices.insert(dev)
 
+        def add_disable_legacy(devices, dev, dev_type):
+            """
+            This function is used to add disable_legacy option for virtio-pci
+            """
+            options = devices.execute_qemu("-device %s,?" % dev_type)
+            if "disable-legacy" in options:
+                value = params.get("disable_legacy", "off")
+                dev.set_param("disable-legacy", value)
+
+        def add_disable_modern(devices, dev, dev_type):
+            """
+            This function is used to add disable_modern option for virtio-pci
+            """
+            options = devices.execute_qemu("-device %s,?" % dev_type)
+            if "disable-modern" in options:
+                value = params.get("disable_modern", "on")
+                dev.set_param("disable-modern", value)
+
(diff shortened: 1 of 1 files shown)


spcui:
ID: 1241333

CongSmile:
Also need set the filter in guest side, as not all guests support this param.

virttest/qemu_vm.py
```diff
@@ -1119,6 +1119,18 @@ def add_balloon(devices, devid=None, bus=None, use_old_format=None):
                 dev.set_param("id", devid)
             devices.insert(dev)
 
+        def add_disable_legacy(devices, dev, dev_type):
+            options = devices.execute_qemu("-device %s,?" % dev_type)
+            if "disable-legacy" in options:
+                value = params.get("disable_legacy", "off")
+                dev.set_param("disable-legacy", value)
```
CongSmile: How about if there is no such option "disable-legacy" ?

virttest/qemu_vm.py
```diff
@@ -1958,6 +1976,15 @@ def add_balloon(devices, devid=None, bus=None, use_old_format=None):
...
                         use_old_format=use_ofmt)
 
+        for device in devices:
+            virtio_pci_devices = ["virtio-net-pci", "virtio-blk-pci",
+                                  "virtio-scsi-pci", "virtio-balloon-pci",
+                                  "virtio-serial-pci", "virtio-rng-pci"]
+            dev_type = device.get_param("driver")
+            if dev_type in virtio_pci_devices:
+                add_disable_legacy(devices, device, dev_type)
+                add_disable_modern(devices, device, dev_type)
```
CongSmile: If the param set "disable-modern = off", but there is no such option "disable-modern" in qemu, which means the qemu does not support such option.  
There should give a warn or case failed directly, which may cause the real problem in testing.

suqinhuang: In my opinion, it's ok without the warning.

suqinhuang:
Acked-by: Suqin Huang

CongSmile:

xutian:
LGTM

merged. thanks @spcui  @CongSmile  and @suqinhuang

PR #2247, opened 2015-07-10
Fix the issue param 'timeout' to float type

vipmike007:
The timeout passed is str type ,can will cause Type Error when i
execute a plus operation.
This patch is to modify timeout as float type.

BZ: 1241372


PandaWei:
>>> timeout=str(1.3)
>>> time.time() + timeout
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for +: 'float' and 'str'


ACK

PandaWei:
Thanks @vipmike007  
Merge Now.

PR #2255, opened 2015-07-15
using notepad.exe for copy insetead of cmd.exe

vipmike007:
When running multi_disk job ,we will hit "The process cannot access t he file because it is being used by another process" when trying to copy cmd.exe from D:\ to C:\

This patch is using notepad.exe to fix it.

ID: 1243345


xutian:
LGTM,

xiagao:
This patch only change file which is used to copy from c: to d: , then recopy it from d: to c:. And confirm C:\WINDOWS\system32\notepad.exe exists in every windowns by now .

Acked-by: Xiaoling Gao <xiagao@redhat.com>

xutian:
Merged, thanks @vipmike007  and @xiagao

PR #2257, opened 2015-07-19
utils_test: increase timeout on md5sum since it can take a while

rbbratta:
For some reason my autotest tarball was huge, so the md5sum
did not complete in the default timeout of 60 seconds

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -699,7 +699,9 @@ def run_autotest(vm, session, control_path, timeout,
         hash_differs = False
         local_hash = utils.hash_file(local_path)
         basename = os.path.basename(local_path)
-        output = session.cmd_output("md5sum %s" % remote_path)
+        output = session.cmd_output("md5sum %s" % remote_path,
+                                    timeout=int(
+                                        params.get("md5sum_timeout", 240)))
         if "such file" in output:
             remote_hash = "0"
         elif output:


xutian:
@rbbratta   Please exclude useless files when make tarball, it will save your test time to make tarball and check md5sum.

@xutian

rbbratta:
Excluding files would require changing the autotest tarball code depending on various conditions.

I don't think that is much justification for an arbitrary 60 second timeout.  Timeouts should only be for recovering from network failures or such things. 

Instead you are arbitrarily cutting off all shell commands after 60 seconds.  Why?

xutian:
In most cases the default timeout(60s) is enough.  If network failures(or such things) happen during your test, I think you'd better to analysis you test log to ensure it's not a bug of test project, not only wait it recover automatically. 
In this case,  you said you autotest tarball is huge, why?  In my test,  default timeout is enough. no failure happen at make autotest tarball step and  check md5sum step.

thanks,
@xutian

rbbratta:
The default timeout is not enough and is arbitrarily low.  If commands complete is less than 60 seconds then the timeout isn't hit so the timeout doesn't save you any time.  It does not slow done the system.   Thus if the timeout is 240 seconds and all your commands complete in less than the timeout it also does not slow down the system.

Here are just a few of the cases in utils_test/__init__.py where the default timeout of 60 seconds is not sufficient.

```python


def stop_windows_service(session, service, timeout=120):
--


def start_windows_service(session, service, timeout=120):
--
        mv_cmd = "/bin/mv %s %s" % (dst_path, error_dst_path)
        utils.system(mv_cmd, timeout=360, ignore_status=True)
--

    def run_host_cmd(host_cmd, timeout=720):
--
        guest_file_create_cmd = "dd if=/dev/zero of=%s bs=1M count=%d"
        guest_file_create_cmd = params.get("guest_file_create_cmd",
                                           guest_file_create_cmd)
        cmd = guest_file_create_cmd % (guest_data_file, count)
        error.context("Creating %dMB file on host" % filesize, logging.info)
        session.cmd(cmd, timeout=600)
--
        logging.debug("Extracting %s on VM %s", basename, vm.name)
        session.cmd("rm -rf %s" % dest_dir, timeout=240)
        dirname = os.path.dirname(remote_path)
        session.cmd("cd %s" % dirname)
        session.cmd("mkdir -p %s" % os.path.dirname(dest_dir))
        e_cmd = "tar xjvmf %s -C %s" % (basename, os.path.dirname(dest_dir))
        output = session.cmd(e_cmd, timeout=240)
--
        logging.debug("Trying to copy autotest results from guest")
        res_index = get_last_guest_results_index()
        guest_results_dir = os.path.join(outputdir, "guest_autotest_results%s" % (res_index + 1))
        os.mkdir(guest_results_dir)
        # result info tarball to host result dir
        session = vm.wait_for_login(timeout=360)
--
        results_tarball = "/tmp/results.tgz"
        compress_cmd = "cd %s && " % results_dir
        compress_cmd += "tar cjvf %s ./*" % results_tarball
        compress_cmd += " --exclude=*core*"
        compress_cmd += " --exclude=*crash*"
        session.cmd(compress_cmd, timeout=600)
        vm.copy_files_from(results_tarball, guest_results_dir)
        # cleanup autotest subprocess which not terminated, change PWD to
        # avoid current connection kill by fuser command;
        clean_cmd = "cd /tmp && fuser -k %s" % results_dir
        session.sendline(clean_cmd)
        session.cmd("rm -f %s" % results_tarball, timeout=240)
--
```

xutian:
Ok, make it configurable is acceptable. 

Thanks,
@xutian

virttest/utils_test/__init__.py
```diff
@@ -699,7 +699,8 @@ def copy_if_hash_differs(vm, local_path, remote_path):
         hash_differs = False
         local_hash = utils.hash_file(local_path)
         basename = os.path.basename(local_path)
-        output = session.cmd_output("md5sum %s" % remote_path)
+        output = session.cmd_output("md5sum %s" % remote_path,
+                                    timeout=params.get("md5sum_timeout", 240))
```
xutian: we should convert it to float or int,  params.get("md5sum_timeout") will return a string if we set it in configuration file.

PandaWei:
LGTM

PandaWei:
thanks @rbbratta and @xutian 
Merge now.

PR #2258, opened 2015-07-19
utils_test: extract tarball when the remote dir not exist

rbbratta:
For this case we had cached the tarball in the VM from a previous run, but we had removed the /usr/local/autotest directory to cleanup.
In this case we don't need a new tarball, we just need to extract the existing tarball if the directory doesn't exit.

### DIFF
--- a/virttest/utils_test/__init__.py
+++ b/virttest/utils_test/__init__.py
@@ -687,6 +687,9 @@ def run_autotest(vm, session, control_path, timeout,
     The following params is used by the migration
     :param params: Test params used in the migration test
     """
+    def directory_exists(remote_path):
+        return session.cmd_status("test -d %s" % remote_path) == 0
+
     def copy_if_hash_differs(vm, local_path, remote_path):
         """
         Copy a file to a guest if it doesn't exist or if its MD5sum differs.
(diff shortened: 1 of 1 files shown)


virttest/utils_test/__init__.py
```diff
@@ -686,6 +686,10 @@ def run_autotest(vm, session, control_path, timeout,
     The following params is used by the migration
     :param params: Test params used in the migration test
     """
+    def directory_exists(vm, remote_path):
+        status, output = session.cmd_status_output("test -d %s" % remote_path)
+        return status == 0
+
```
PandaWei: Hi @rbbratta 
you don't seem to care the ouput.
so session.cmd_status() is enough, and you can return it directly.

will-Do:
ACK

virttest/utils_test/__init__.py
```diff
@@ -686,6 +686,10 @@ def run_autotest(vm, session, control_path, timeout,
     The following params is used by the migration
     :param params: Test params used in the migration test
     """
+    def directory_exists(vm, remote_path):
```
xutian: why vm param used ?

rbbratta: removed unused arg and submitted new version

xutian:
@rbbratta 
 Your IDE make value of some configuration to wrong, please reconfigure your IDE then format your patch.

Thanks,
Xu

rbbratta:
I can't find anything in the diff, is it whitespace?

xutian:
@rbbratta, sorry add comment to wrong PR.

PandaWei:
Thanks all the people who participated in this discussion.
Merge now.

PR #2259, opened 2015-07-20
virttest: small enhancement for get_cpu_count

xutian:
sometimes,  we got unprintable string in console output, convert a string to int will cause exception, so use regex to filter number string only.


xutian:
ID:  1069147

spcui:
LGTM
Acked-by: scui@redhat.com

sathnaga:
Acked-by: sathnaga@linux.vnet.ibm.com

PR #2265, opened 2015-07-23
Provide portal_ip for iscsi setup.

rbian:
Ipv6 portal address can'be created by default, we need to add a parameter
add then create ipv6 portal if needed.

### DIFF
--- a/virttest/iscsi.py
+++ b/virttest/iscsi.py
@@ -11,10 +11,10 @@ iscsi in localhost then access it.
 import re
 import os
 import logging
-import ipaddr
 from autotest.client import os_dep
 from autotest.client.shared import utils, error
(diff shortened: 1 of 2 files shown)


waynesun09:
ipaddr not python standard library, so travis failed unittest at import it.

rbian:
@waynesun09 , Yes, any suggestion to avoid this?

waynesun09:
@rbian 
since ipaddress is in python 3, so we can't use it also.
then maybe wirte a new function in utils_net using socket

waynesun09:
ACK

waynesun09:
@rbian I'm merging this now, looking forward your fix on lun_num on iscsi, thx

PR #2266, opened 2015-07-23
Add iothread option.

rbian:


rbian:
Duplicated with 2265, close this one.

PR #2269, opened 2015-07-27
Disable ppc e1000

vipmike007:
e1000 & rtl8139 is not support in ppc 
this patch is to disable them on power platform

ID: 1247467

### DIFF
--- a/shared/cfg/guest-hw.cfg
+++ b/shared/cfg/guest-hw.cfg
@@ -3,14 +3,19 @@
 # NICs
 variants:
     - rtl8139:
+        no ppc64 ppc64le
         nic_model = rtl8139
     - e1000:
+        no ppc64 ppc64le
         nic_model = e1000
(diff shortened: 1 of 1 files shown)


xutian:
Currently, spapr-vio bus doesn't support hot-pluging, so variant spapr-vlan shouldn't be add to nic_hotplug test too.

xutian:
so you still need a patch in tp-qemu repo

vipmike007:
@xutian yes ,you are right. https://github.com/autotest/tp-qemu/pull/423

vipmike007:
After talking w/ ppc feature owner ,this patch is not related to 1214221.I will use 1247467 instead to track it.

xutian:
LGTM,

CongSmile:

xutian:
Merged thanks @vipmike007  and @CongSmile

PR #2271, opened 2015-07-29
virttest.virsh: Bug fix for event()/qemu_monitor_event()

Chenditang:
In qemu_monitor_event() function, the parameter of timeout should be 
event_timeout, not event.
event() function has the same problem.

### DIFF
--- a/virttest/virsh.py
+++ b/virttest/virsh.py
@@ -4094,7 +4094,7 @@ def qemu_monitor_event(domain=None, event=None, event_timeout=None,
     if event:
         cmd += " --event %s" % event
     if event_timeout:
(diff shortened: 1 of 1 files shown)


PandaWei:
Code is OK for me.
But i noticed that there's no enough information and explanation in your commit(s),
including these commits had been merged.
Please explain what's the bug in commit log and update this PR.

PandaWei:
And please notice the length of each line for commit log

PandaWei:
@Chenditang
I means that you should update your commit log, not just only the conversation.
Each commit should have detailed log information.

will-Do:
Fair enough.
ACK and merge it

PR #2274, opened 2015-07-29
Enable USB type device passthrough using add_hostdev within virt test

bssrikanth:
Currently hostdev addition supported adding only pci type device for 
passthrough. This commit has added capability to within virt test 
framework to generate a hostdev xml section for USB device passthrough. 
Guest VM xml add hostdev will now be able accept USB specific parameters 
and add respective xml clause to guest VM accordingly.


bssrikanth:
Sorry will run pep8 compliance locally.. fix issues and then create pull request.

PR #2282, opened 2015-08-05
virttest.qemu_vm: add workaround for ppc console issue

luckyh:
For old kernels (< 3.1), here is a console issue - if the very last
byte of the spapr-vty's "reg" property is not equal to zero, then it
will not appeared as 'hvc0' which is expected. So made a workaround
here to slove this problem. Details:
    http://lists.gnu.org/archive/html/qemu-ppc/2013-10/msg00129.html


luckyh:
ID: 1239070

xutian:
LGTM,

PR #2284, opened 2015-08-06
iscsi: avoid aggressive clearconfig

waynesun09:
As multiple iscsi target could be created, use clearconfig to clear
entire local configuration will cause problem, so after each delete
use saveconfig will be safer.


PandaWei:
LGTM

waynesun09:
Thx @PandaWei 
merge this as trivial fix.

PR #2288, opened 2015-08-07
Fix skip message for TestFail()

PandaWei:
In order to display message to users correctly,
The exception message has to use interpolation.
This PR refers to the commit beacb9d.

Signed-off-by: Wei,Jiangang <weijg.fnst@cn.fujitsu.com>


lmr:
Looks good, thanks!

PR #2291, opened 2015-08-17
virt-test: base.cfg: add ehci as the default usb device controller.

CongSmile:
Signed-off-by: Cong Li <coli@redhat.com>

id: 1183906

### DIFF
--- a/shared/cfg/base.cfg
+++ b/shared/cfg/base.cfg
@@ -85,10 +85,10 @@ cdroms = cd1
 usbs = usb1
 # USB controller type, run following command to see supported controller.
 # `qemu-kvm -device \? 2>&1 | grep "usb.*bus PCI"`
-usb_type = ich9-usb-uhci1
-usb_type_usb1 = ich9-usb-uhci1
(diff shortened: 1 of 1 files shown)


CongSmile:
Based on tp-qemu/pull/442

CongSmile:
Close this PR, will do it in internal.

shared/cfg/base.cfg
```diff
@@ -85,10 +85,10 @@ cdroms = cd1
 usbs = usb1
 # USB controller type, run following command to see supported controller.
 # `qemu-kvm -device \? 2>&1 | grep "usb.*bus PCI"`
-usb_type = ich9-usb-uhci1
-usb_type_usb1 = ich9-usb-uhci1
```
suqinhuang: I'd like to re-open this patch,
and 
and usb_controller = uhci 
and remove 
"usb_controller_tablet1 = uhci" ==> so it will works even more devices are added
"usb_type = ich9-usb-uhci1" ==> useless

PR #2293, opened 2015-08-17
Return lun number in setup_or_cleanup_iscsi.

rbian:
Lun number is needed to build domain xml for iscsi disk, store
the lun number during iscsi setup and make it available to iscsi object.

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -577,7 +577,7 @@ def define_pool(pool_name, pool_type, pool_target, cleanup_flag, **kwargs):
         extra = "--source-name %s" % vg_name
     elif pool_type == "iscsi":
         # Set up iscsi target without login
-        iscsi_target = setup_or_cleanup_iscsi(True, False)
+        iscsi_target, _ = setup_or_cleanup_iscsi(True, False)
         cleanup_iscsi = True
         extra = "--source-host %s  --source-dev %s" % ('localhost',
                                                        iscsi_target)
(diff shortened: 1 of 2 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -474,7 +474,7 @@ def setup_or_cleanup_iscsi(is_setup, is_login=True,
             utils.run("rm -f %s" % emulated_path)
         else:
             _iscsi.export_target()
-            return emulated_target
+            return (emulated_target, _iscsi.luns)
```
waynesun09: With this update, all setup_or_cleanup_iscsi with is_setup=True and is_login=False should be updated.
Please checking following code and also update tp-libvirt side, Thanks.

rbian: Yes, I have checked all tests, please refer to autotest/tp-libvirt#604

waynesun09: I think you need to check following here:
virttest/utils_test/libvirt.py:580
virttest/utils_test/libvirt.py:2030

waynesun09:
ACK

PandaWei:
@waynesun09 and @rbian 
In my opinion,
It's not a good idea that return (emulated_target, _iscsi.luns) .
Not all of the caller of setup_or_cleanup_iscsi() needs the lun number.
And this PR involves a lot of functions.
we should try our best to control the scope of the changes.

We can add function named get_luns() by emulated_target or  emulated_target at the class of _IscsiComm .
and call it when we need it.

rbian:
@PandaWei , I have thought about it. 
As we can get the lun number during the initialization time, the easiest way is to return the value by setup_or_cleanup_iscsi.
If we add new mothed get_luns to the class of _IscsiComm, we must have a Iscsi object to call the function, the object can be seen in setup_or_cleanup_iscsi, and it's not in tp-libvirt. We also can't get lun number except return the Iscsi object to tp-libvirt.

PandaWei:
@rbian 
Yes， No Iscsi object is a problem.
and It seems that there's no good idea to avoid it so far.
I'm OK for this PR.

waynesun09:
Thanks @rbian @PandaWei , merge now.

PR #2294, opened 2015-08-20
virttest.utils_test.libvirt: Missed iothread option.

rbian:
Missed iothread option for disk paramters in set_vm_disk

### DIFF
--- a/virttest/utils_test/libvirt.py
+++ b/virttest/utils_test/libvirt.py
@@ -1958,6 +1958,7 @@ def set_vm_disk(vm, params, tmp_dir=None, test=None):
     emu_image = params.get("emulated_image", "emulated-iscsi")
     image_size = params.get("image_size", "10G")
     disk_format = params.get("disk_format", "qcow2")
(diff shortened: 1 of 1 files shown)


virttest/utils_test/libvirt.py
```diff
@@ -1958,6 +1958,7 @@ def set_vm_disk(vm, params, tmp_dir=None, test=None):
     emu_image = params.get("emulated_image", "emulated-iscsi")
     image_size = params.get("image_size", "10G")
     disk_format = params.get("disk_format", "qcow2")
+    driver_iothread = params.get("driver_iothread")
```
waynesun09: better give a default value other than None, or xml parsing will have problem

rbian:
It's okay to set a None value, the value will be checked later:
        if driver_iothread:
            driver_attrs['iothread'] = driver_iothread
Anyway, I will use null string "" instead.

waynesun09:
@rbian ok, i see. 
Thanks, merge as trivial fix.

PR #2296, opened 2015-08-24
utils_v2v: Remove the obsolete '-f' option

will-Do:
Remove the '-f' option as it already discarded. And rename the variable
of input mode to avoid conflict with build-in function name.


waynesun09:
ACK

waynesun09:
thx and merge now.

